PR #50, opened 2023-02-16
Fixed missing word in README.md

Sayamame-beans:
Fixed missing word.

PR #51, opened 2023-02-16
docs: insert line break in command description of readme

Sayamame-beans:
I think it might be better to do this.
If not, reject this PR.
depends on: #50


anatawa12:
I think it's not necessary unless one line is too long.

PR #299, opened 2023-12-29
Introduce pending project changes

anatawa12:
Behavior Changes:

- `vrc-get resolve <package> <version>` is now invalid. it's not expected
- `vrc-get resolve --prerelease` is now invalid. it's not expected
- `vrc-get remove` now confirms user if removing package conflicts with existing packages. (previously errors)

### DIFF
--- a/vrc-get-vpm/src/environment.rs
+++ b/vrc-get-vpm/src/environment.rs
@@ -1,8 +1,8 @@
+mod empty;
 mod repo_holder;
 mod repo_source;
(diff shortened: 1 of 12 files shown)


anatawa12:
TODO: 動作確認

anatawa12:
- `vrc-get resolve`
- `vrc-get install`
- `vrc-get remove`
- `vrc-get upgrade`

動作確認よし

PR #329, opened 2024-01-16
fix: 1.5.0 is dynamic linked

anatawa12:


anatawa12:
bypass since it's not related to code, build config

PR #527, opened 2024-03-05
Fix some phrases and grammatical errors

lu-na-r:
I've fixed some grammatical errors in README. I've also changed some words and phrases so that readers can easily understand what you want to say.


anatawa12 (approved):
LGTM. Thank you

Issue #703, opened 2024-04-01
"Settings" title card is different from other title cards


lu-na-r:
### Description of the bug / バグの説明

Title cards for "Projects", "VPM Repositories" and "Logs" are bold text on gray card. Contrary to those, "Settings" title card is thin text on white card.

### Steps to Reproduce / 再現手順

Open each menu.

### Expected Behavior / 想定される挙動

The title card on "Setting" menu should be like the other style, or vice versa (I think thin text on white card is better)

### Screenshots / スクリーンショット

"Logs" title card
<img width="749" alt="スクリーンショット 2024-04-01 22 38 12" src="https://github.com/vrc-get/vrc-get/assets/100114898/597ecfa6-fbaa-4de3-9d50-8cefa61fe45d">

"Settings" title card (this is neater than the other)
<img width="759" alt="スクリーンショット 2024-04-01 22 38 24" src="https://github.com/vrc-get/vrc-get/assets/100114898/b01b492b-9e4d-42e5-9bd7-355af9399a1f">

### Software Version / バージョン情報

v0.1.0-beta.17

### OS

macOS 14.3.1

PR #721, opened 2024-04-06
Fixes issue #703 Settings Title card is different

Spokeek:
Fixed it by using the VStack and HNavBar layout components to match the other pages design

Fixes https://github.com/vrc-get/vrc-get/issues/703


anatawa12 (approved):
Thank you for contribution!

LGTM

Issue #671, opened 2024-03-27
Bottom side of project list card is not masked properly


lu-na-r:
### Description of the bug / バグの説明

The project list card is round-shaped, but the bottom side is not masked, which makes the card weird.

### Steps to Reproduce / 再現手順

Just open the app.

### Expected Behavior / 想定される挙動

The card should be masked so that it becomes round-shaped.

### Screenshots / スクリーンショット

Top
<img width="790" alt="image" src="https://github.com/vrc-get/vrc-get/assets/100114898/965062df-a3e3-4a68-9227-cb5d6cd1703e">

Bottom
<img width="790" alt="スクリーンショット 2024-03-27 14 33 58" src="https://github.com/vrc-get/vrc-get/assets/100114898/8ca4efc4-9a5a-42a3-b487-198c028e8f7f">

### Software Version / バージョン情報

0.1.0-beta.15

### OS

macOS 14.3.1

PR #722, opened 2024-04-06
Fix issue #671 Project List corder design

Spokeek:
Fixed the issue that was caused by the shadow effect of the card wrapping around the table

Fix https://github.com/vrc-get/vrc-get/issues/671


anatawa12 (approved):
LGTM Thank you!

Issue #716, opened 2024-04-05
[style] 設定ページ内で、説明文と題目がほぼ同じ表示で見づらさを感じるので差を付けて欲しい


Sayamame-beans:
### Summary / 概要

タイトル通りです。現在のPR #714 で内容が増えてより強く感じる気がします。
(大きさ等に差が無いので題目と説明という分かれ方に見えない?)
![image](https://github.com/vrc-get/vrc-get/assets/61457993/31153b0c-dbdb-49d0-924b-e61108c1a804)
![image](https://github.com/vrc-get/vrc-get/assets/61457993/03a1c978-0ab7-42f2-b432-8c76a9d15674)

### Purpose / 目的

見やすくなる

anatawa12:
あれ？ここもウェイトの差があるはずなんだけど...
<img width="680" alt="image" src="https://github.com/vrc-get/vrc-get/assets/22656849/ed155f7b-3f70-4777-9700-40a6fbe23c65">

Sayamame-beans:
うぇ…
しばらく前から差が付いてなかったような気がする? (i18n実装する前には付いてたのかもしれませんが記憶していない)

Spokeek:
I could get my hands into this. I would see if I can harmonize the settings page as a whole so each section uses the same design and so we avoid inconsistencies.

Spokeek:
The original issue comes from tailwind that disables all browser specific design values to only give power to css classes that you apply to components.

So here i can re enable the sizing values in the css globals.
By also adding additional margins so the sections are clearer, would give something like this 

![image](https://github.com/vrc-get/vrc-get/assets/4519853/7386e119-5beb-4cd3-930c-d7394e828fc2)

Tell me if that's a good idea or if i should remove all header values and use the classes instead.

anatawa12:
> Tell me if that's a good idea or if i should remove all header values and use the classes instead.

 I think it's ok to use  css globals.

PR #731, opened 2024-04-06
Cleaned layout for Backup section in Settings page

Spokeek:
Fixes https://github.com/vrc-get/vrc-get/issues/716


anatawa12:
image

vrc-get-gui/app/settings/page.tsx
```diff
@@ -225,14 +225,17 @@ function Settings(
...
+					<h3>{tc("backup path")}</h3>
+					<Typography className={"whitespace-normal"}>
+						{tc("the backup path is the directory where vrc-get-gui will create backup zips of the projects.")}
+					</Typography>
+					<div className={"flex gap-1"}>
+						<Input className="flex-auto" value={settings.project_backup_path} disabled/>
+						<Button className={"flex-none px-4"} onClick={selectProjectBackupFolder}>{tc("select")}</Button>
+					</div>
 				</div>
+				<div className="mt-2">
```
anatawa12, line 238: please add indent

anatawa12 (changes requested):
I think the new

vrc-get-gui/app/settings/page.tsx
Spokeek: I'm not sure i understand what you mean.

This class adds two units of margin top to the component so it can be further from the section title.

What would you want to do here ?

Spokeek: Did you mean to add some padding left so the text can be pushed to the right ?

anatawa12: please add indent to line 239-247

vrc-get-gui/app/settings/page.tsx
```diff
@@ -225,14 +225,17 @@ function Settings(
...
 				<label className={"flex items-center"}>
```
anatawa12, line 239:
```suggestion
				    <label className={"flex items-center"}>
```

Something like this

Sayamame-beans (approved):
This has resolved the problem to some extent. Looks good!

(off-topic: I noticed that the area behind the `:` has no space, but this is not caused by this PR, but was there originally. (out of the area of this PR))

vrc-get-gui/app/settings/page.tsx
Spokeek: I don't know how i missed that and didn't understand :/ sorry

anatawa12 (approved):
Thank you!

PR #736, opened 2024-04-06
Chore: Using common Component to structure Settings page

Spokeek:
As i was working on the settings page, i saw we had common components in the layout file.

I followed that use and placed a Section Component that wraps our entries in Cards for our various pages.

For now as a proposal i did the **settings** page and the **logs** as they were the more useful for.

This is **only a proposal** on how we can do our sections so we can keep the design homogeneous.

Tell me what you think about it.

image

### DIFF
--- a/vrc-get-gui/app/log/page.tsx
+++ b/vrc-get-gui/app/log/page.tsx
@@ -64,6 +64,26 @@ export default function Page() {
 }
 
 function logEntryToText(entry: LogEntry) {
-	return `${entry.time} [${entry.level.padStart(5, ' ')}] ${entry.target}: ${entry.message}`;
+	const level = entry.level
+	let levelClassName;
+	switch(level){
+		case "Debug":
+			levelClassName = "bg-blue-600"
+			break
+		case "Info":
+			levelClassName = "bg-green-600"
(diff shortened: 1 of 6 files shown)


anatawa12:
I think creating common components is excellent. Thank you for proposal.

I think sections should be in vertical, and path input and button should be grouped in another horizontal component.

Spokeek:
> I think sections should be in vertical, and path input and button should be grouped in another horizontal component.

I'll try to make this. Would it be so we have a seperate component for each horizontal group ? Like for the example of the two backup sections ?

anatawa12:
I think having common component for path and pick pair and put inside VSection component is good.

Spokeek:
I Reworked it again, making use of the material Typography component.

Trying to follow the Component based structure.
May try to generalize the UI Components as well like the Table.

image
image
image

Spokeek:
As i used my components for the log page, i also added a color hint to clearer the log level of each line.
image

Spokeek:
And applied the table component to repository list 
image

anatawa12:
When will this pr become ready to merge? I want to merge this and implement color theme

PR #786, opened 2024-04-15
chore(gui): create .nvmrc

enitimeago:
Example usage:

```
vrc-get-gui % fnm use
Can't find an installed Node version matching v20.x.x.
Do you want to install it? answer [y/N]: y
Installing Node v20.12.2 (arm64)
Using Node v20.12.2
```


anatawa12:
I mean I think the developer can use either v18 or v20 at your choice depending on the installed / currently using version. I think it's unnessesary to force to use v20.

anatawa12:
> Should the README be changed to describe this?

Oh sorry, I didn't think much at that time. I think it should be changed.

PR #789, opened 2024-04-15
Fix CHANGELOG-gui.md

lu-na-r:
Fixed some typos in CHANGELOG-gui.md.

* "localisation" is not wrong in Britain, but you've used American spelling in every context, so I fixed it.


anatawa12:
Thank you

PR #812, opened 2024-04-19
feat(vrc-get-vpm): add `/opt/unityhub/unityhub` for detection candidate

KisaragiEffective:
This commit allows the detection method to include binary via [their official](https://docs.unity3d.com/hub/manual/InstallHub.html#install-hub-linux) APT distribution.


anatawa12 (approved):
LGTM. Thank you!

PR #876, opened 2024-04-29
German translation: Add "projects:type:sdk2"

JustBuddy:
Adds new key `projects:type:sdk2` from #869 to the german translation file.

SDK2 doesn't need translation, so this is just for completions-sake.

Ready to merge.


anatawa12 (approved):
Thank you!

Issue #895, opened 2024-05-01
Need for french translation ?


Spokeek:
### Summary / 概要

Well, title is self explanatory. I'm French native so i can do that translation. But i'm unsure if there is a need for it.

### Purpose / 目的

More appeal for French users.

anatawa12:
I think having more localization will increase merits of ALCOM so adding localization is very welcome!

Spokeek:
> I think having more localization will increase merits of ALCOM so adding localization is very welcome!

Feel free to add a discussion if required, i'll base my work on the english translation then.

anatawa12:
Because of design in my workflow, I could not add to the action unless we have so could you create pr first?

Spokeek:
Done @anatawa12

PR #904, opened 2024-05-02
French Translation

Spokeek:
Relates to #895


Spokeek:
I didn't do a 1o1 translation, i especially kept a few words that aren't really used in french language. Like logs, package, things like that.
also replaced "pre release packages" by something like "experimental packages", as pre release sounds quite weird in french.

If we have some more feedback about it, i can always change it.

anatawa12:
> I didn't do a 1o1 translation, i especially kept a few words that aren't really used in french language. Like logs, package, things like that.

I think keeping technical words in english is good if there is no good words in French. 
The term "prerelease" is also a technical word so it might be good to keep "prerelease" instead of using other words but using "experimental" is also good.

PR #911, opened 2024-05-05
chore(l10n/ja): follow the updates

Sayamame-beans:
for https://github.com/vrc-get/vrc-get/discussions/855#discussioncomment-9317872


anatawa12 (commented):
Thank you!

PR #925, opened 2024-05-07
Fix EN localisation

lu-na-r:
I put period in the end of some sentences and fixed grammar mistakes.

Detail:
- https://github.com/vrc-get/vrc-get/discussions/894#discussioncomment-9294328
- You can not use to-infinitive with "recommend"


anatawa12 (approved):
LGTM! Thank you!

PR #936, opened 2024-05-11
feat(gui): Separate Unity Hub source locales key

lonelyicer:
Separate the ‘Unity Hub’ of the path settings from the ‘Unity Hub’ of the installation source.


anatawa12 (approved):
LGTM. Thank you

PR #970, opened 2024-05-13
chore(gui): renamed zh_CN to zh_Hans

lonelyicer:
Make the language naming style the same as Traditional Chinese


anatawa12:
zh-CN means Chinese language for China and zh-Hans means Chinese language with Simplified Chinese Ideographs.

It's good to rename. however we have to do more to rename. do you want to those? or should I do?
- change default language name for `zh-` locale to `zh_hans` in `config.rs`
- write migration in `config.rs` (if `language` in config is `zh_cn`, replace with `zh_hans`)
- update locale id in `localization-updates.js`

lonelyicer:
> zh-CN means Chinese language for China and zh-Hans means Chinese language with Simplified Chinese Ideographs.
> 
> It's good to rename. however we have to do more to rename. do you want to those? or should I do?
> 
> * change default language name for `zh-` locale to `zh_hans` in `config.rs`
> * write migration in `config.rs` (if `language` in config is `zh_cn`, replace with `zh_hans`)
> * update locale id in `localization-updates.js`

done

anatawa12:
it looks config migration is not implemented yet.

lonelyicer:
> it looks config migration is not implemented yet.

config.rs has been changed. Is there anywhere else?🤔

anatawa12:
> Is there anywhere else?

yes. we have to care old uses who has `zh_cn` in config file. after reading file, replace `zh_cn` with `zh_hans`.

lonelyicer:
image


It looks like there's no place where `zh_cn` still exists.

anatawa12:
we have to add some code to support `config.json` in user environment. not on source code.

lonelyicer:
Oh, I see what you mean, the user still has zh_cn locally and needs to write something to replace the user's local zh_cn with zh_hans.

anatawa12:
Yes.

lonelyicer:
> we have to add some code

But I don’t know how to write `Rust`, so I can only trouble you.😖

PR #973, opened 2024-05-14
"Custom" to "Custom Template"

lu-na-r:
https://github.com/vrc-get/vrc-get/discussions/894#discussioncomment-9422107


anatawa12:
Thank you!

PR #984, opened 2024-05-17
feat(gui): change project manage page layout

lonelyicer:
Before:
image

After:
image

### DIFF
--- a/vrc-get-gui/app/projects/manage/page.tsx
+++ b/vrc-get-gui/app/projects/manage/page.tsx
@@ -528,28 +528,6 @@ function PageBody() {
 
 						<SearchBox className={"w-max flex-grow"} value={search} onChange={e => setSearch(e.target.value)}/>
 
-						{packageRows.some(row => row.latest.status === "upgradable") &&
-							<Button className={"flex-shrink-0"}
-											onClick={onUpgradeAllRequest}
-											disabled={isLoading}
-											color={"green"}>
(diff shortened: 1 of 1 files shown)


anatawa12:
Umm...
It's reasonable to rearrange to put `…` to the left but I feel `Select Repositories` button should be on the left

How do you think `Upgrade All` `Select Repositories` `…` order?

lonelyicer:
> How do you think `Upgrade All` `Select Repositories` `…` order? or putting some divier (space) between `Select Repositories` and `…` with original order?

Putting ‘...’ to the far right of the layout is more intuitive to the user, because it means ‘more action’.
Other apps generally do the same.

lonelyicer:
As for `Upgrade All` and `Select Repositories`, I think the order from left to right should be `Select Repositories` `Upgrade All`,  because the actions given in `...` are related to the VPM Package, and it would be more intuitive to put the buttons that are more functionally related together.
However, this is not the most important part, and perhaps their positions could be switched.

anatawa12:
After discussing with Sayamame-beans and some other friends yesterday night, I decided to keep current UI for now.

Thank you for suggestion!

Sayamame-beans:
Just a note...
Upgrade All button won't be displayed if there's nothing to update, so if we place it in the center or to the right, other buttons' position will change depending on whether there are some upgradable or not.

PR #998, opened 2024-05-19
feat(gui): localization package source

lonelyicer:
image


anatawa12:
Thank you for suggestion

I think this should be implemented after #628 and we have to consider if we should inject into VRChat official repository or not so unfortunately I don't merge for now. thank you for suggestion.

For future me: if we choose to inject into vrchat official repository, I think we should detect repository by repository id instead of name.

PR #999, opened 2024-05-19
ci: use github app token for labeler

### LINKED ISSUE #867: CI: changelog check should be executed after Labeler
Labeler adds labels for the pr but changelog check will not be executed after that so I should reconfigure to execute changelog check after labeler

anatawa12:
fixes #867


anatawa12:
testing with #1000

PR #1002, opened 2024-05-19
chore(i18n): add unity version info for unity not found toast

Sayamame-beans:
I would like to add unity version info to unity not found toast like #988 dialog.


anatawa12 (commented):
よさそう

PR #1053, opened 2024-05-29
chore(i18n/de): add keys for #1046

JustBuddy:
Adds german translation for:
```
projects:manage:toast:the package has newer latest with incompatible unity
projects:manage:toast:some package has newer latest with incompatible unity
projects:manage:tooltip:incompatible with unity
```


anatawa12:
Thank you!

PR #1108, opened 2024-06-11
chore(i18n/de): Improved translations.

JustBuddy:
Improved some wording so it works better in the german language.


anatawa12:
Thank you!

Issue #1129, opened 2024-06-19
use single icon library?


anatawa12:
Currently, ALCOM uses heroicons (as recommended in material tailwind) and lucide icons (as shadcn uses) so I think we should replace heroicons with lucide if possible

PR #1173, opened 2024-06-25
chore: Move from heroicons to lucide-react

JustBuddy:
Fixes https://github.com/vrc-get/vrc-get/issues/1129

What do you think?
image
image

From my test, it is safe to uninstall the heroicons package.
Please do so if you end up using this, I am not sure if I did the uninstallation correctly.

### DIFF
--- a/CHANGELOG-gui.md
+++ b/CHANGELOG-gui.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog].
 - Refine Dark Theme `#1083`
 - Show package display name on uninstalled toast `#1086`
 - Improved performance of package search `#1131`
+- Swapped heroicons for lucide-react `#1129`
 
 ### Deprecated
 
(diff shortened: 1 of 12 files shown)


anatawa12:
Thank you for implementation!


It looks good in general but the star icon on the checkbox is less noticeable I think so could you increase stroke width of star icon on the favorite checkbox?

https://lucide.dev/guide/basics/stroke-width

> From my test, it is safe to uninstall the heroicons package.

I don't like to create another PR to remove the heroicons from package.json so could you remove it from package.json?

JustBuddy:
> It looks good in general but the star icon on the checkbox is less noticeable I think so could you increase stroke width of star icon on the favorite checkbox?
> 
> https://lucide.dev/guide/basics/stroke-width

Sure, I'll look into that.

> I don't like to create another PR to remove the heroicons from package.json so could you remove it from package.json?  

Do I just remove it from package.json? That's all?

JustBuddy:
That's how different widths look.
imageimageimage
default - **width 3** - width 4

I think 3 is best, 4 looks a bit odd.

anatawa12:
> Do I just remove it from package.json? That's all?

I think we need to call npm install again to update package-lock.json also.

JustBuddy:
done, should be good now.

anatawa12 (approved):
LGTM! Thank you!

PR #1262, opened 2024-07-14
use Biome for lint and formatting

anatawa12:


anatawa12:
todo: exclude biome from bindings.ts

.git-blame-ignore-revs
```diff
@@ -0,0 +1 @@
+798da7111ab611c07ff7171574cd2ab9e8254515 # reformat with biomejs
```
anatawa12, line 1:
```suggestion
# reformat with biomejs
798da7111ab611c07ff7171574cd2ab9e8254515
```

anatawa12:
多分大丈夫やろ

PR #1370, opened 2024-08-10
fix: sometimes fails to load litedb due to opening same file multiple time

anatawa12:


vrc-get-vpm/src/io/tokio.rs
```diff
@@ -67,13 +67,32 @@ impl EnvironmentIo for DefaultEnvironmentIo {
...
+        let lock = {
+            use sha1::Digest;
+
+            let path_lower = path.to_lowercase();
+            let mut sha1 = sha1::Sha1::new();
+            sha1.update(path_lower.as_bytes());
+            let hash = &sha1.finalize()[..];
+            let hash_hex = hex::encode(hash);
+            let name = format!("Global\\{hash_hex}.Mutex");
+
```
anatawa12, line 84:
```suggestion


```

vrc-get-vpm/src/io/tokio.rs
```diff
@@ -67,13 +67,32 @@ impl EnvironmentIo for DefaultEnvironmentIo {
...
+        #[cfg(windows)]
+        let lock = {
+            use sha1::Digest;
+
+            let path_lower = path.to_lowercase();
+            let mut sha1 = sha1::Sha1::new();
+            sha1.update(path_lower.as_bytes());
+            let hash = &sha1.finalize()[..];
+            let hash_hex = hex::encode(hash);
+            let name = format!("Global\\{hash_hex}.Mutex");
```
anatawa12, line 83:
```suggestion
            // this lock name is same as shared engine in litedb
            let name = format!("Global\\{hash_hex}.Mutex");
```

PR #1397, opened 2024-08-14
Updated French Translations from 0.1.14

Spokeek:
Removed non existing keys

Added the missing ones with french translations.

If required the translations can be reordered.


anatawa12:
Thank you~

> If required the translations can be reordered.

Key ordering is completely up for maintainer of each locale so reordering is not required.

Sayamame-beans:
> If required the translations can be reordered.

not required, but I'll do that if you want.

PR #1494, opened 2024-09-08
French translation

Spokeek:
Here is the french translation required for the next release.

There is a lot of changes as now i make use of a script to add missing keys automatically (can be found here https://gist.github.com/Spokeek/f7e478d7c293f04c295943252ca7f941)


JustBuddy:
Translation keys at the moment are sorted by category and appearance, the script seems to just rearrange them to alphabetical order. Is that correct?

I like our current approach as it is rather easy to work with.

JustBuddy:
What you submitted doesn't follow the current layout, I think you comitted the scripted output.

Also some keys seem broken:
image

Spokeek:
> What you submitted doesn't follow the current layout, I think you comitted the scripted output.
> 
> Also some keys seem broken, or don't follow the json formatting.: image

thats what i thought too, but we dont follow json but json5 format. i used the same lib used in the translation script next to it.

also pushing the updated version is the plan

Spokeek:
> Translation keys at the moment are sorted by category and appearance, the script seems to just rearrange them to alphabetical order. Is that correct?
> 
> I like our current approach as it is rather easy to work with.

and yes, the workflow is described in the script, but we can discuss it , i made a discussion for it https://github.com/vrc-get/vrc-get/discussions/1495

CHANGELOG-gui.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog].
 ### Added
 
 ### Changed
+- Updated French Translation [`#1442`](https://github.com/vrc-get/vrc-get/pull/1494)
```
anatawa12, line 13: localization changes doesn't need changelog. (CI was confused because helper script is added(

.github/scripts/load-missing-localization-keys.js
anatawa12, line 1: Adding shbang and make the file executable might be better

anatawa12 (commented):
Thank you for updating the localization!

I don't like to use sorting but it's your locale so I'll merge them.

Spokeek:
> Thank you for updating the localization!
> 
> I don't like to use sorting but it's your locale so I'll merge them.

I don't want to go all on my way and differentiate with the other locales. I'll see what to do on the discussion before merging if this is fine

CHANGELOG-gui.md
Spokeek: I reverted the commit.

Spokeek: Will potentially do a separate MR for the script

.github/scripts/load-missing-localization-keys.js
Spokeek: Added

Spokeek:
Damn, broken linux build ... no idea what causes that.

JustBuddy:
> Damn, broken linux build ... no idea what causes that.

Probably just randomly failed, let @anatawa12 rerun the checks.

Spokeek:
I updated it all. @anatawa12 ready when you are

anatawa12:
Thank you!

Issue #1332, opened 2024-07-29
ALCOM Compatibility With Flatpak Unity Hub


CorneliusCornbread:
### Description of the bug / バグの説明

During the setup of ALCOM it doesn't detect that I have the flatpak version of the Unity Hub installed, it seems to not be satisfied until I point it to an executable file so pointing it to either of the executables in the .ld.so (in ~/.var/app/com.unity.UnityHub/) folder let me continue but doesn't actually work for launching Unity itself. Attempting to open a project just results in it saying that I don't have a valid active Unity license.

### Steps to Reproduce / 再現手順

1. Install flatpak Unity Hub
2. Open ALCOM
3. ALCOM complains of being unable to find Unity Hub
4. Point to the only executable within the app folder

### Expected Behavior / 想定される挙動

It'd be nice if it could automatically detect the flatpak install, otherwise some documentation on what to point to for the flatpak install would also work just fine.

### Software Version / バージョン情報

v0.1.13

### OS

Fedora 40 KDE Plasma Spin

anatawa12:
license problem might be related to #653

For unity hub path detection, could you let us know the absolute path of unity hub executable installed with flatpak?

CorneliusCornbread:
> license problem might be related to #653
> 
> For unity hub path detection, could you let us know the absolute path of unity hub executable installed with flatpak?

I genuinely have no clue which executable in the flatpak is the unity executable I'll be honest.

anatawa12:
I think good to support flatpak installation of unity hub l however I currently doesn’t have enough time for researching flatpack and flatpack distribution of Unity Hub so I need more information or help for fixing this problem.

ProffDea:
> For unity hub path detection, could you let us know the absolute path of unity hub executable installed with flatpak?

`/var/lib/flatpak/app/com.unity.UnityHub/current/active/files/extra/unityhub`

`current` and `active` are both symlinks. The real path to the flatpak app's **folder** without symlinks can be found with the following command `flatpak info -l com.unity.UnityHub` which will output something like:

- `/var/lib/flatpak/app/com.unity.UnityHub/x86_64/stable/174c4c04a73215d38a16aba984fd9c414be0a9c33801b3eb536357badc590f47`

`files/extra/unityhub` will need to be appended to the output of the command to actually launch unity hub.

```bash
#!/bin/sh
# This will launch unity hub
$(flatpak info -l com.unity.UnityHub)/files/extra/unityhub
```

[...]

PR #1551, opened 2024-09-26
Added support for Flatpak version of Unity Hub

gizmo-ds:
This adds support for the Flatpak version of Unity Hub. It works fine with Unity Hub 3.9.1 (Flatpak).

image


gizmo-ds:
I'm not sure if adding this checkbox under the Unity Hub Path is the best approach, but I couldn't think of a better way to handle it.

anatawa12 (changes requested):
Thank you for creating code for support flatpak!

However, I don't like to use `flatpak` as the special value for flatpak support and I don't want to include special support with `flatpak` command, I want to support with adding  `/var/lib/flatpak/app/com.unity.UnityHub/current/active/files/extra/unityhub` as default unity hub search path list

gizmo-ds:
> Thank you for creating code for support flatpak!
> 
> However, I don't like to use `flatpak` as the special value for flatpak support and I don't want to include special support with `flatpak` command, I want to support with adding `/var/lib/flatpak/app/com.unity.UnityHub/current/active/files/extra/unityhub` as default unity hub search path list as I said in #1332.
> 
> Is there any other reason to implement this way?

Thanks for the feedback! I understand the preference for directly executing the Unity Hub executable. However, doing so would bypass the advantages of the Flatpak sandbox environment. I believe there might be other ways to approach this while still retaining the benefits of the sandbox.

anatawa12:
umm.... I thought `/var/lib/flatpak/app/com.unity.UnityHub/current/active/files/extra/unityhub` is wrapper executable.

PR #1585, opened 2024-10-04
chore(l10n): [de] update locale

JustBuddy:
Added china-specific language keys.
As I can't realistically test this, these are provided to the best of my abilities without seeing the actual dialogues.


anatawa12:
Thank you!

PR #1586, opened 2024-10-05
feat: add flatpak installation of unity hub to default search path

### LINKED ISSUE #1332: ALCOM Compatibility With Flatpak Unity Hub

anatawa12:
Close #1332

Waiting flathub/com.unity.UnityHub#126


vrc-get-vpm/src/environment/unity_management.rs
```diff
@@ -211,8 +211,15 @@ fn default_unity_hub_path() -> &'static [&'static str] {
...
-            static ref INSTALLATIONS: [&'static str; 3] =
-                [&USER_INSTALLATION, "/usr/bin/unity-hub", "/opt/unityhub/unityhub"];
+            static ref INSTALLATIONS: [&'static str; 4] =
+                [
+                    &USER_INSTALLATION,
+                    "/usr/bin/unity-hub",
+                    // apt package
+                    "/opt/unityhub/unityhub",
+                    // flatpak
+                    "/var/lib/flatpak/export/bin/com.unity.UnityHub",
```
ProffDea, line 221: I made one easy mistake in my last comment referring to the path of unityhub, the word `export` is suppose to be plural for system wide installed flatpaks, so the correct path is `exports`.
`/var/lib/flatpak/exports/bin/com.unity.UnityHub`

ProffDea:
I've learned today that flatpaks can be installed in the user's home directory rather than being system wide. Unity hub can potentially be installed at:
- `$HOME/.local/share/flatpak/exports/bin/com.unity.UnityHub`

vrc-get-vpm/src/environment/unity_management.rs
anatawa12:
```suggestion
                    "/var/lib/flatpak/exports/bin/com.unity.UnityHub",
```

PR #1913, opened 2025-02-14
Remove uppercase from Button component

guilty-vrc:
Forcing button text to be converted to uppercase may lead to unexpected wording.
The conversion to uppercase should be specified in each value within locale files.
Therefore, in this PR, I have removed the uppercase conversion from the Button component.
Additionally, I searched for any other instances of uppercase, lowercase, or capitalize being used, but found none, so there should be no issues.

### DIFF
--- a/vrc-get-gui/components/ui/button.tsx
+++ b/vrc-get-gui/components/ui/button.tsx
@@ -10,21 +10,21 @@ const buttonVariants = cva(
 		variants: {
 			variant: {
 				default:
-					"bg-primary text-primary-foreground hover:bg-primary/90 shadow-primary/50 hover:shadow-primary/50 shadow hover:shadow-md transition-shadow uppercase",
+					"bg-primary text-primary-foreground hover:bg-primary/90 shadow-primary/50 hover:shadow-primary/50 shadow hover:shadow-md transition-shadow",
 				destructive:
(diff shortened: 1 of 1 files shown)


anatawa12:
Thank you for proposing change. showing button as uppercase is expected to me. Could you let me know which place is unexpected?

anatawa12:
For settings.json and gui-config.json, I feel it's not good to uppercase. you're correct.

however, it's expected for most other places.
Uppercase is came from shadcn component system and I think it's good (at least not bad) so I keep as uppercase.

anatawa12:
Sorry I thought it was came from shadcn but it was actually came from [material tailwind](https://www.material-tailwind.com/docs/react/button) the component library I initially implemented with. (I left material tailwind with due to some problems and lack of features)

> vrc-get-gui/components/ui/button.tsx is a generic button component.
For that reason, I believe enforcing uppercase would be a bad practice.

When we prefer uppercase by default (no decision are made yet), I feel it's better to set uppercase by default in generic component, optionally disable with additional css is better than enabling uppercase everywhere or manually uppercase for ALL components in locale file.

>  locale files are written in Xxxx Xxxx style,

It's just my decision in my locale time big long time ago so no big reason to keep in Xxx Xxxx in en.json

anatawa12:
> I apologize if I'm repeating myself, but if this is simply your preference, there is absolutely no issue with the current implementation.

I want to discuss pros and cons before making decision.
Sorry for taking time for discussion.

anatawa12:
原則UPPERCASEのままにする方針を維持します

PR #1964, opened 2025-03-06
chore(l10n): [de_DE] update locale and line-break fix for en.

JustBuddy:
The englisch template uses \n for a new string instead of <br/>, resulting in that line break being ignored.

Also new strings for german locale.


anatawa12:
thank you for fix and quick update.

PR #2043, opened 2025-03-28
several frontend small migrations

anatawa12:


anatawa12:
> Ref forwarding has been removed from TooltipContent. Confirm that external ref usage is no longer required or reinstate forwardRef if needed.

False positive. ref forwarding is performed with `...props`. React 19 can do this.

Copilot, Please understand React 19 has ref

PR #2047, opened 2025-03-30
chore(I10n): [ko_KR] update locale

CirnoV:


anatawa12:
Thank you for update and sorry for late response!

PR #2056, opened 2025-03-31
chore(deps): bump deps

anatawa12:


anatawa12:
it looks there is breaking changes in tauri cli.

it looks `-c '{"version":"${{ steps.version.outputs.version }}", "bundle":{"windows":{"certificateThumbprint": null}}}'` does not work well, `certificateThumbprint` does not become null.

anatawa12:
tauri-apps/tauri#13114

downgrade tauri-cli and proceed for now

PR #2103, opened 2025-04-23
fix: log page auto scroll is broken

lonelyicer:


lonelyicer:
Also idk why my biome is broken
I try to run `npm run biome -- check --fix` and than I got a lot of errors
image

anatawa12 (commented):
Thank you for fixing issue!

However I feel it's not correct to nest `requestAnimationFrame`.
I read nesting `requestAnimationFrame` is to delay frame for 2 frames, but browser implementation may alter over time to requiring 3 or only 1 frames.
Is there any resources on web about nesting `requestAnimationFrame`?

lonelyicer:
> However I feel it's not correct to nest `requestAnimationFrame`.

Regarding this part, my original intention was to ensure the DOM updates are completed before handling the scroll. But perhaps nesting requestAnimationFrame is indeed unnecessary.

vrc-get-gui/app/_main/log/-logs-list-card.tsx
```diff
@@ -27,23 +25,34 @@ export const LogsListCard = memo(function LogsListCard({
...
+			if (!scrollContainerRef.current) return;
+
+			const container = scrollContainerRef.current;
+			const isNearBottom =
+				container.scrollHeight -
+					(container.scrollTop + container.clientHeight) <
+				50;
+
+			if (!isNearBottom) {
+				container.scrollTop = container.scrollHeight;
```
anatawa12, line 44: AS far as I tried and according to react's docs, it looks requestAnimationFrame is not necessary (useEffect will be called after DOM update) so can you try remove wrapping with requestAnimationFrame?

anatawa12 (approved):
LGTM. Thank you!

PR #2114, opened 2025-04-28
fix: template_dir doesn't exists

lonelyicer:


vrc-get-gui/src/commands/environment/templates.rs
```diff
@@ -200,6 +200,9 @@ async fn save_template_file(
 
     let mut file = 'create_file: {
         let template_dir = Path::new("vrc-get/templates");
+        if !io.is_dir(template_dir).await {
+            io.create_dir_all(template_dir).await?;
+        }
```
anatawa12, line 205:
```suggestion
        io.create_dir_all(template_dir).await?;
```

create_dir_all doesn't throw any error if already exists so checking for directory can be removed.

anatawa12 (changes requested):
Thank you for fixing issue.

Can you add PR number 2114 to `New Project Template System` line in CHANGELOG-gui.md please?

anatawa12 (approved):
LGTM! Thank you

PR #2147, opened 2025-05-09
Enhanced Template Builder and Backend Improvements

hardtokidnap:
I have been working on some enhancements for `vrc-get` as a hobby programmer, and your project has been a great source of inspiration and learning for me. 

I wanted to contribute something back to the community and also expand my own understanding of Rust and frontend development.

This pull request introduces several changes, primarily focused on improving the template functionality and addressing a couple bugs i tripped on. 

I hope you can use some if not all of it. If you don't take in my changes, that's okay, I hope it respects the quality you want. 

A note on the .gitignore, cursor seems to demand that these are added, even though my rules for that damned AI is empty. Trade-off for autocompletion. This autocompletion also seemed to completely fuck up your Markdown and the json5, doing changes i didn't ask for. I stopped trying to combat it.

However, here it goes:

Key changes include:

* **Enhanced template builder:**
    * Introduced more robust handling for custom project templates.
    * Improved the logic for resolving template dependencies, including merging VPM package versions and Unity packages from base and custom templates.
    * Enhanced the existing functionality for creating and saving user defined templates.
    * Included a new frontend component for selecting multiple packages.
    * Added a shortcut to open the custom template directory for easier user access.

* **Backend modifications:**
    * During the development of the template features, I encountered an issue with how Unity versions were parsed, especially for custom templates that might use formats like `2022.3.6f1-EXTRA`. To ensure these could be handled correctly, I updated the Unity version parsing logic within the `vrc-get-vpm` crate. This change allows for more flexible and accurate parsing of Unity versions.
    * Improved template ID generation and added validation to reserve specific namespaces while allowing user created ones.
    * Enhanced logging in the backend for template processing to aid in troubleshooting.

* **Frontend and general updates:**
    * Made several UI adjustments in the project creation section to better support the new template features and improve user experience.
    * Updated localization files with new strings for the added features.
    * Added changelog entries for `vrc-get-gui` and initiated a changelog for `vrc-get-vpm` to document the Unity version parsing improvement. (used #references as commit ID)
    * The `.gitignore` file was updated.
    * I tried to re-use some of your components, but i hit a few snags and had to do changes outside of my own files - I apologise beforehand if i made any mistakes and broke something else in the progress. I have QA'd and it didn't seem like it, but you'll never know.

I tried to work around the existing backend systems as much as possible, but the bug related to Unity version parsing required modifications to ensure custom templates could function reliably. I believe these changes will make the template system more powerful and user friendly.

[...]

PR #2163, opened 2025-05-13
feat(backend): Implement stable and preserved IDs for user templates

hardtokidnap:
From [this PR](https://github.com/vrc-get/vrc-get/pull/2147#pullrequestreview-2830837961)

Preserving template id for user templates.

### DIFF
--- a/vrc-get-gui/src/commands/environment/templates.rs
+++ b/vrc-get-gui/src/commands/environment/templates.rs
@@ -123,13 +123,19 @@ pub async fn environment_save_template(
     vpm_packages: Vec<(String, String)>,
     unity_packages: Vec<String>,
 ) -> Result<(), RustError> {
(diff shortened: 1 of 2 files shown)


anatawa12 (changes requested):
Preserving id would help in some case but I don't want to allow specifying `com.anatawa12.vrc-get.user.` templates as a base template therefore could you add verification on parsing templates that denies `com.anatawa12.vrc-get.user.` as base template.

Previous version prevented this by assigning random id for non-user-id-assigned user templates, but this PR would assign random id at creation so I want to have check for `com.anatawa12.vrc-get.user.` instead.

Being inherited by other templates requires some extra care to not break something so I want to prevent for normal templates.

hardtokidnap:
Closing, no longer valid.

PR #2164, opened 2025-05-13
Feat(Backend): Improve custom template resolution and Unity version filtering

hardtokidnap:
[Addressing this PR](https://github.com/vrc-get/vrc-get/pull/2147#pullrequestreview-2830837961), request for separate PR's and trimmed logging.

### DIFF
--- a/vrc-get-gui/src/templates.rs
+++ b/vrc-get-gui/src/templates.rs
@@ -190,29 +190,47 @@ pub async fn load_resolve_alcom_templates(
             let alcom = template.alcom_template.as_ref().unwrap();
             let Some(base) = template_by_id.get(&alcom.base) else {
                 // The template will never become available so remove from keys to update
+                log::debug!("Template {}: Base template {} not found.", k, alcom.base);
                 return false;
             };
 
             if !base.available {
                 // The base template is not available yet. Retry later
+                log::debug!("Template {}: Base template {} not available yet.", k, alcom.base);
                 return true;
             }
 
-            // The base template is available! update this template based on the base template
+            let mut unity_versions = Vec::new(); // Start with empty
 
-            let unity_versions = if let Some(unity_filter) = &alcom.unity_version {
-                base.unity_versions
-                    .iter()
-                    .copied()
-                    .filter(|x| unity_filter.matches(&x.as_semver()))
-                    .collect()
+            if let Some(unity_filter) = &alcom.unity_version {
+                let unity_filter_str = unity_filter.to_string();
+                match vrc_get_vpm::version::UnityVersion::parse(&unity_filter_str) {
+                    Some(specific_version) => {
+                        log::debug!("  Parsed as specific version: {}", specific_version);
+                        if base.unity_versions.contains(&specific_version) {
+                            log::debug!("   Specific version is compatible with base.");
+                            unity_versions.push(specific_version); // Use only this specific version
+                        } else {
+                            log::warn!("   Specific version {} is NOT compatible with base versions: {:?}", specific_version, base.unity_versions);
+                        }
+                    }
+                    None => {
+                        unity_versions = base.unity_versions
+                            .iter()
+                            .copied()
+                            .filter(|x| unity_filter.matches(&x.as_semver()))
+                            .collect();
+                    }
+                }
             } else {
-                base.unity_versions.clone()
-            };
+                unity_versions = base.unity_versions.clone()
+            }
 
             let template_mut = &mut template_by_id[k];
             template_mut.unity_versions = unity_versions;
-            template_mut.available = true;
+            // Mark as available only if we found at least one compatible Unity version.
+            template_mut.available = !template_mut.unity_versions.is_empty();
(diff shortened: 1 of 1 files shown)


vrc-get-gui/src/templates.rs
```diff
@@ -190,29 +190,47 @@ pub async fn load_resolve_alcom_templates(
...
+                            .copied()
+                            .filter(|x| unity_filter.matches(&x.as_semver()))
+                            .collect();
+                    }
+                }
             } else {
-                base.unity_versions.clone()
-            };
+                unity_versions = base.unity_versions.clone()
+            }
```
anatawa12, line 227: What's the purpose of this change?

The version name `2022.3.22` would only match `2022.3.22f1` therefore previous logic work as expected.

The `f1` part is much less important than `2022.3.22` part and due to underlying library limitation, I choose not supporting specifying `f1` part.
If there is popular useful use case, we should change the `Range` struct to support unity style version number instead I think.

vrc-get-gui/src/templates.rs
```diff
@@ -472,22 +490,39 @@ pub async fn create_project(
...
                         Entry::Occupied(mut e) => {
-                            let range = range.intersect(e.get());
-                            e.insert(range);
+                            log::debug!(
+                                "Merging existing dependency: {} (Base: {}, Custom: {})",
+                                pkg_id,
+                                e.get(),
+                                range
+                            );
+                            let intersected_range = range.intersect(e.get());
```
anatawa12, line 507: It might be better (and enough) to log intersected_range and not log base / custom since those templates are likely to be already resolved and logged before.
```suggestion
                            let intersected_range = range.intersect(e.get());
                            log::debug!(
                                "Merged existing dependency: {} {}",
                                pkg_id,
                                intersected_range,
                            );
```

anatawa12 (commented):
I think it's good to add debug logs but I have one question about implementation.

vrc-get-gui/src/templates.rs
anatawa12: Why did you resolved this question without answering?

vrc-get-gui/src/templates.rs
anatawa12: Also this. no response

vrc-get-gui/src/templates.rs
hardtokidnap: Apologies, these should have been addressed, not sure why it wasn't posted. I commented directly from VS. 

I'll get back to you

vrc-get-gui/src/templates.rs
hardtokidnap: As mentioned, these should have been addressed via VS, but they somehow didn't post. I'll get back to you

vrc-get-gui/src/templates.rs
hardtokidnap: I’ve reverted the change to the original logic, as the previous approach was based on a different system on the frontend.

vrc-get-gui/src/templates.rs
hardtokidnap: I've revisited and will recommit with the suggested changes. An oversight on my end since it was a part of a debugging i did.

vrc-get-gui/src/templates.rs
anatawa12: You said 'reverted' but you're chanigng in this PR. you may forgot to push some commits?

hardtokidnap: There we go :)

vrc-get-gui/src/templates.rs
hardtokidnap: Updated.

vrc-get-gui/src/templates.rs
anatawa12: Do you understand github's comment system?

I'm commenting and asking for line 205..227 but you have only reverted 226..227.

hardtokidnap: Sorry for the continued mistakes and confusion >.< My github extension has had a stroke and it's pushing comments that i've made long after they're supposed to be commented.

It should now be updated with the requested changes. Again, my bad. I am not used to cooperating with others.

vrc-get-gui/src/templates.rs
```diff
@@ -472,22 +476,36 @@ pub async fn create_project(
...
+                            log::debug!(
+                                "Merged existing dependency: {} {}",
+                                pkg_id,
+                                intersected_range,
+                            );
+
```
anatawa12, line 499: missingle multiple `}`s.

anatawa12 (changes requested):
Please build and run formatter in your enviroment before pushing please>?

vrc-get-gui/src/templates.rs
anatawa12: Don't copy and paste AI results please?

Yes you have to insert if vacant and update intersected range if occupied.

3rd suggestion looks correct suggestion but unrelated to this PR.

PR #2165, opened 2025-05-13
Allow parsing Unity version format like 2022.3.22-f1

hardtokidnap:
From [this PR](https://github.com/vrc-get/vrc-get/pull/2147#pullrequestreview-2830837961), review suggestion to split into it's own PR.

### DIFF
--- a/vrc-get-vpm/src/version/unity_version.rs
+++ b/vrc-get-vpm/src/version/unity_version.rs
@@ -78,19 +78,37 @@ impl UnityVersion {
         }
     }
 
-    // expects major.minor.revision[type]increment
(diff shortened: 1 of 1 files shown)


anatawa12 (commented):
As I described before, `2022.3.222-f1` format is not valid unity version format so we should avoid generating such format instead of supporting them I think.

anatawa12:
And In addition please follow [conventionalcommits](https://www.conventionalcommits.org/en/v1.0.0/) style commit messages please? Sorry for my fault that I forgot to mention in the CONTRIBUTING.md

hardtokidnap:
Closing this PR to re-visit with a better approach to ensure only standard formats are generated.

PR #2167, opened 2025-05-13
feat(gui): Implement package selection in template creation dialog

hardtokidnap:
From [this pr ](https://github.com/vrc-get/vrc-get/pull/2147#pullrequestreview-2830837961)

Implement package selection in template creation dialog and fix bug to unity package clipping with `overflow-y-auto`


vrc-get-gui/app/_main/packages/templates/index.tsx
```diff
@@ -569,108 +658,65 @@ function TemplateEditor({
...
+												<SelectValue className={"grow"} />
+											</SelectTrigger>
+											<SelectContent>
+												{[...new Set(installedUnityVersions)].map((version) => (
+													<SelectItem key={version} value={version}>
+														{version}
+													</SelectItem>
+												))}
+											</SelectContent>
+										</Select>
```
anatawa12, line 672: Could you use something like https://v0.dev/chat/fork-of-combobox-demo-WYfVsfS4QT2 with https://ui.shadcn.com/docs/components/combobox?
This allows both filly customize version range and user's selection

vrc-get-gui/components/PackageMultiSelect.tsx
anatawa12, line 1: could you add version row? and I think description row is not necessary 

This card would also can use something like https://v0.dev/chat/fork-of-combobox-demo-WYfVsfS4QT2 with ui similar to previous version

vrc-get-gui/locales/en.json5
```diff
@@ -141,6 +141,7 @@
     "projects:button:create": "Create",
     "projects:creating project...": "Creating the project...",
     "projects:toast:project created": "Project was created successfully.",
+    "templates:dialog:unitypackage path": "Unity package path",
```
anatawa12, line 144: Please add to area with many "templates"

vrc-get-gui/app/_main/packages/templates/index.tsx
```diff
@@ -475,16 +548,39 @@ function TemplateEditor({
...
 
+	// Parse installed unity versions as version strings (e.g., '2022.3.22f1') from the unity_paths
+	const installedUnityVersions = unityVersionsResult.data?.unity_paths?.map(([, version]) => version) || [];
+
+	// Auto-select unity version if only one is installed, or select the first available if none is selected
+	React.useEffect(() => {
+		if (installedUnityVersions.length > 0 && (!unityRange || !installedUnityVersions.includes(unityRange))) {
+			setUnityRange(installedUnityVersions[0] || '');
+		}
+	// eslint-disable-next-line react-hooks/exhaustive-deps
```
anatawa12, line 559: we're using biome as a linter & formatter. we're not using eslint

vrc-get-gui/app/_main/packages/templates/index.tsx
```diff
@@ -436,24 +451,82 @@ function TemplateEditor({
...
+				id !== "com.vrchat.core.vpm-resolver"
+			));
+		}
+	}, [baseTemplate]);
+
+	// Fetch available packages and repository info
+	const information = useSuspenseQuery(environmentProjectCreationInformation);
+	const packagesResult = useSuspenseQuery(environmentPackages);
+	const repositoriesInfo = useSuspenseQuery(environmentRepositoriesInfo);
+	const unityVersionsResult = useSuspenseQuery(environmentUnityVersions);
```
anatawa12, line 516: use `useSuspenceQueries` to make them request parallelly.
When we use `useSuspenseQuery` in a row, it will be a request waterfall problem

vrc-get-gui/app/_main/packages/templates/index.tsx
```diff
@@ -436,24 +451,82 @@ function TemplateEditor({
...
+				id !== "com.vrchat.avatars" &&
+				id !== "com.vrchat.worlds" &&
+				id !== "com.vrchat.base" &&
+				id !== "com.vrchat.core.vpm-resolver"
+			));
+		}
+	}, [baseTemplate]);
+
+	// Fetch available packages and repository info
+	const information = useSuspenseQuery(environmentProjectCreationInformation);
```
anatawa12, line 513: it looks information is not used

vrc-get-gui/app/_main/packages/templates/index.tsx
```diff
@@ -436,24 +451,82 @@ function TemplateEditor({
...
+			});
+		} else if (baseTemplate === "com.anatawa12.vrc-get.blank") {
+			setSelectedPackages((prev) => prev.filter(
+				id =>
+				id !== "com.vrchat.avatars" &&
+				id !== "com.vrchat.worlds" &&
+				id !== "com.vrchat.base" &&
+				id !== "com.vrchat.core.vpm-resolver"
+			));
+		}
```
anatawa12, line 509: on the new template layer, I think it's not necessary to check base packages.

In addition, I think it's not good to un-check packages when changed to blank template

vrc-get-gui/components/PackageMultiSelect.tsx
anatawa12, line 1: This component is specific to templates page so please add to `app/_main/packages/templates/-package-multi-select.tsx`

vrc-get-gui/app/_main/packages/templates/index.tsx
hardtokidnap: @anatawa12 I don't see the issue with un-checking. If you select a template (e.g. Avatars) it _should_ both check and highlight these at the top. When switching back to a blank template, it _should_ revert to the blank one. If we retain the selection from Avatars onto the blank template, this would look like a bug or a lazy job in my eyes.

We might be talking across from eachother here?

- The precise definition of "blank": Should it merely not add any packages, or should it actively remove a specific set of packages to ensure a clean slate regarding VRChat specific items?

- The user workflow: What is the anticipated sequence of actions? If a user selects "Avatars", then switches to "Blank", is the goal usually to start fresh, or to slightly modify the "Avatars" setup under a new "custom" guise?

Should we perhaps:

- ONLY change the packages that would be included with VRChat when swapping between the different base templates and retain the custom packages that are chosen?
  - This ensures that even if the user has a custom selection but wants to switch from avatars to world, it would retain the custom selection and only change the base template?

vrc-get-gui/components/PackageMultiSelect.tsx
hardtokidnap: @anatawa12 What specifically are you pointing to here?

vrc-get-gui/app/_main/packages/templates/index.tsx
hardtokidnap: @anatawa12, thanks for the Combobox suggestion

Following your feedback, I've implemented a working prototype of the Combobox on the frontend.
This new Combobox allows users to either type a custom Unity version string (e.g., a specific version like `2022.3.6f1` or a range like `>=2019.4`) or select from the list of installed Unity versions. The frontend state now correctly captures this flexible input and is ready to pass it to the backend.

The immediate next steps involve backend changes to ensure this `unityRange` string (which can be a specific version or a range) is properly handled by `environmentSaveTemplate`:

* The Rust command `environmentSaveTemplate` needs to be adapted to accept this string.
* This string then needs to be validated and parsed into a `vrc_get_vpm::version::VersionRange` on the backend.
* The resulting `VersionRange` (or its string representation) will be stored in the `.alcomtemplate` file.
* Finally, `load_resolve_alcom_templates` needs to correctly load and utilize this string/range for template availability checks.

As you can see, the full functionality of the enhanced Combobox (allowing custom ranges) is tightly coupled with these backend modifications. Pushing just the frontend Combobox change without the corresponding backend logic to save and load these new string/range formats would lead to a broken template saving/loading. Reverting parts of the existing backend logic from the `template-builder` branch to try and isolate just the Combobox UI also would just create a cluster of faults.

Additionally, there are other review comments (related to Rust backend logic for template ID generation, base template validation, logging, VCC template export, etc.) that are part of this larger effort to enhance the template building feature.

Considering these interdependencies, I believe a PR that includes the frontend Combobox, its necessary backend support for handling the Unity version string, and addresses the other related review comments from the patch would form a more coherent and testable functional unit.

Would you be comfortable with me proceeding to integrate these backend changes and address the other comments, and then submitting a more comprehensive PR? Or do you have a specific preference for how these interconnected pieces should be broken down further? I want to ensure the changes are reviewable effectively while maintaining the integrity of the feature.

anatawa12: > This new Combobox allows users to either type a custom Unity version string (e.g., a specific version like 2022.3.6f1 or a range like >=2019.4) or select from the list of installed Unity versions.
> 
> - The Rust command environmentSaveTemplate needs to be adapted to accept this string.
> - This string then needs to be validated and parsed into a vrc_get_vpm::version::VersionRange on the backend.
> - The resulting VersionRange (or its string representation) will be stored in the .alcomtemplate file.
> - Finally, load_resolve_alcom_templates needs to correctly load and utilize this string/range for template availability checks.

We don't have to change them since it already accept both `>=2019.4` and `2022.3.6`. As I described on https://github.com/vrc-get/vrc-get/pull/2167#discussion_r2087086796 and https://github.com/vrc-get/vrc-get/pull/2164#discussion_r2086973032, It's not planned to support `2022.3.22f1` style of unity version.

For front-end validation of version name, we can use `rangeRegex` in templates/index.tsx.

vrc-get-gui/components/PackageMultiSelect.tsx
anatawa12: This file is not generic component used from many place, this component is used only by templates page so I want to put templates page and component specifically for templates together.
When I choose to use this component in other place, it's the time move to components directory.

vrc-get-gui/app/_main/packages/templates/index.tsx
anatawa12: The custom template should only specify additional packages so not adding VRCSDK to the list is the correct way I think

hardtokidnap: For the current implementation i have done this (I have been working all day, just ironing out a few bugs)

When you choose a "Base template", the following happens:

- Base template
  - Take away any VRCSDK packages
  - Keep selection of custom packages
- Avatars
  - Add vrchat.avatars, vrchat.base, vrchat.vpm-resolver
- World
  - Add vrchat.worlds, vrchat.vpm-resolver, vrchat.base

This ensures everybody can use this feature.. See the quick demo i set up here of the nearly finished feature.

https://streamable.com/1nrxuh

What's left is handling edgecases where unitypackage is missing meta files (Handle error, continue but warn user) and fine-tuning.

anatawa12: I mean don't check VRChat SDK on user changed base template to VRCSDK Worlds/Avatars since those packages are specified in base layer, not on customization layer

hardtokidnap:
All comments should have been taken care of and changelog is updated. Hope this is satisfactory, sorry for the delay, long work week + constitution day :smile: @anatawa12

anatawa12:
It looks base packages are still being selected in template so you're not following my direction.

I'm feeling you may not have proper understanding of design of this feature...
I feel your code lacks some experience and sense of typescript or programming like using some special value that potentially have risk of conflict with original data and using inefficient traditional algorithm...

hardtokidnap:
> It looks base packages are still being selected in template so you're not following my direction.
>
> I'm feeling you may not have proper understanding of design of this feature...
>
> I feel your code lacks some experience and sense of typescript or programming like using some special value that potentially have risk of conflict with original data and using inefficient traditional algorithm...

Excuse the formatting errors if there are any. I'm travelling.

This current version was built from your previous review comments. To say that I am confused about this comment would be an understatement.

Based on earlier review, the following changes were made:

- As per your wish for template specific components, `PackageMultiSelect` was moved to the template specific path and is now imported from `app/_main/packages/templates/-PackageMultiSelect.tsx`.

- **useSuspenseQueries**: Following your instruction to prevent request waterfalls, the `TemplateEditor` was changed to utilize `useSuspenseQueries` for concurrently fetching all necessary environment data.

- **PackageMultiSelect UI**: Based on your feedback for the `PackageMultiSelect` component itself, its internal structure was adjusted, for instance, to include a version row and omit an unnecessary description row, aligning with your comment.

- In line with your comment  
  > Don’t select VRChat SDK on user changed base template… since those packages are specified in base templates, not on customization layer  
  and  
  > Custom template should only specify additional packages

  The `TemplateEditor` UI was changed to filter the list presented in `PackageMultiSelect`. Logic using `implicitPackagesForCurrentBase` and `conflictingSdkPackagesToHide` ensures that users are only shown, and can therefore only select, packages that are genuinely additional to the chosen base template. Base packages are not presented as selectable options in this UI.

- **Linter/Formatter Alignment**: The codebase should align with the project's use of Biome, as per your comment regarding ESLint.

Additionally, the following changes were done across the board:

- Implemented a combobox using the example from your comment.  
  - This also allows regex patterns and validates these.  
  - Allows to set specific Unity version, installed or available versions.  
  - This also happens live, so the user doesn't get an error on saving, losing their template.

- Fixed a bug that didn't allow editing saved templates

This brings me to your latest comment on the current reply:  
**"It looks base packages are still being selected in template so you’re not following my direction."**

To be absolutely clear about my intentions and what I understood from a previous comment: `saveTemplate()` logic in `TemplateEditor`:

a) `userExplicitPackages` are gathered from the user's UI selections. As detailed in point #4 above, this list inherently excludes base SDKs as they are filtered out from direct user interaction in the `PackageMultiSelect` component.

b) The system then iterates through the `BASE_TEMPLATE_PACKAGE_MAP` (which defines inherent packages for each base template). If a package from this map isn't already present in `userExplicitPackages` with a user specified version (which, for base packages, it wouldn't be, due to the UI filtering), it is then merged into the `finalVpmDependencies` list that gets saved, typically with a default version (e.g. "*").

Are you referring to the second step (b) as the specific unwanted behavior? This step involves the systemic merge of these pre defined base packages into the saved template's dependency list, even though they are not user selectable in the UI. If the intended direction is that the `finalVpmDependencies` persisted for a custom template should exclusively contain only the `userExplicitPackages` (from step a), and not include these system merged base packages from `BASE_TEMPLATE_PACKAGE_MAP`, please confirm this specific requirement for the saved data structure. If so, this merge logic will be modified.

Regarding your other general remarks:

> "proper understanding of design"  
> "lack of typescript or programming sense"  
> "inefficient traditional algorithm"

If something stands out as wrong or careless, just point to it. I am eager to learn, as I am far from a professional nor am I used to following others. My experience lies purely in script languages, one pagers and small time python apps.

My goal is for this implementation to keep up the good quality that vrc-get is. If further changes are needed in particular areas of this version, please point that out, and I will address them. If it's too much, I'll leave you to it.

anatawa12:
> Are you referring to the second step (b) as the specific unwanted behavior?

Yes. I told you "The custom template should only specify additional packages" before but still specifying.

> If it's too much, I'll leave you to it.

I feel it's too much for size of change so I'm closing this PR sorry.

PR #2169, opened 2025-05-14
Load projects packages asynchronously

### LINKED ISSUE #1946: Cache project list to reduce load time
### Summary / 概要

I keep my projects on a large hard drive, and it takes quite a while >30sec for the project list to finish loading when ALCOM is opened. I would like to request that the list of projects be cached somewhere so that the project list can be displayed immediately, meanwhile in the background ALCOM should do the usual refreshing process and update the list accordingly when it is done. It is fine of some features that require accurate data are disabled until the actual refresh is complete. But functions such as "Open Project" and "Create Project" should not be blocked.

### Purpose / 目的

This way, the list of projects can be displayed immediately, since it is most likely nothing has changed between runs and all I really want to do is click "Open Project".

### Additional context / 補足

_No response_

anatawa12:
Fixes #1946

Currently projects are only implemented.
We'll implement asynchronous package loading soon.


anatawa12:
Asynchronous package loading has been implemented!

Since my network is very fast and all repositories can use etag cache for loading, It's hard to experience difference but I hope there are improvements

PR #2170, opened 2025-05-15
feat(templates): refine template ID generation and validation

hardtokidnap:
Improves the handling of template IDs in the backend:
- When saving a new template, generates a unique ID within the
  'com.anatawa12.vrc-get.user.' namespace if no ID is provided.
- Enhances ID validation to allow '.user.' and '.vcc.' sub-namespaces
  under 'com.anatawa12.vrc-get.*' while reserving other sub-namespaces.
- Prevents user-defined templates ('.user.*') from being used as
  base templates.

Also fixes a bug with invalid template ID on edit template

### DIFF
--- a/vrc-get-gui/src/templates/alcom_template.rs
+++ b/vrc-get-gui/src/templates/alcom_template.rs
@@ -106,19 +106,32 @@ pub fn parse_alcom_template(json: &[u8]) -> serde_json::Result<AlcomTemplate> {
     let template = serde_json::from_slice::<AlcomTemplateContent>(json)?;
 
     // few validations
-    if let Some(id) = &template.id {
-        if id.0.starts_with("com.anatawa12.vrc-get") {
+    if let Some(id_wrapper) = &template.id { // id_wrapper is &TemplateId
+        // Reserve `com.anatawa12.vrc-get.*` except for known namespaces which are allowed for user-created
+        // templates (".user") and imported VCC templates (".vcc").
+        if id_wrapper.0.starts_with("com.anatawa12.vrc-get")
+            && !(id_wrapper.0.starts_with("com.anatawa12.vrc-get.user.")
+                || id_wrapper.0.starts_with("com.anatawa12.vrc-get.vcc."))
+        {
             return Err(serde_json::Error::invalid_value(
-                Unexpected::Str(&id.0),
-                &"a valid alcom template id (reserved id)",
+                Unexpected::Str(&id_wrapper.0),
+                &"a valid alcom template id (reserved id, or disallowed com.anatawa12.vrc-get.* namespace)",
             ));
         }
+    }
+
+    // Validation for template.base
+    if template.base.0.starts_with("com.anatawa12.vrc-get.user.") {
+        return Err(serde_json::Error::invalid_value(
+            Unexpected::Str(&template.base.0),
+            &"user-defined templates (com.anatawa12.vrc-get.user.*) cannot be used as a base template",
+        ));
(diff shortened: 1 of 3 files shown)


vrc-get-gui/src/templates/alcom_template.rs
```diff
@@ -106,19 +106,32 @@ pub fn parse_alcom_template(json: &[u8]) -> serde_json::Result<AlcomTemplate> {
...
 
     // few validations
-    if let Some(id) = &template.id {
-        if id.0.starts_with("com.anatawa12.vrc-get") {
+    if let Some(id_wrapper) = &template.id { // id_wrapper is &TemplateId
+        // Reserve `com.anatawa12.vrc-get.*` except for known namespaces which are allowed for user-created
+        // templates (".user") and imported VCC templates (".vcc").
+        if id_wrapper.0.starts_with("com.anatawa12.vrc-get")
+            && !(id_wrapper.0.starts_with("com.anatawa12.vrc-get.user.")
+                || id_wrapper.0.starts_with("com.anatawa12.vrc-get.vcc."))
```
anatawa12, line 114: We should deny `com.anatawa12.vrc-get.vcc.` since it's for vcc templates without alcomtemplate file.

vrc-get-gui/src/templates/alcom_template.rs
```diff
@@ -106,19 +106,32 @@ pub fn parse_alcom_template(json: &[u8]) -> serde_json::Result<AlcomTemplate> {
...
+        {
             return Err(serde_json::Error::invalid_value(
-                Unexpected::Str(&id.0),
-                &"a valid alcom template id (reserved id)",
+                Unexpected::Str(&id_wrapper.0),
+                &"a valid alcom template id (reserved id, or disallowed com.anatawa12.vrc-get.* namespace)",
```
anatawa12, line 118: "disallowed com.anatawa12.vrc-get.* namespace" is part of reserved id so we don't have to add message

vrc-get-gui/src/templates/alcom_template.rs
```diff
@@ -106,19 +106,32 @@ pub fn parse_alcom_template(json: &[u8]) -> serde_json::Result<AlcomTemplate> {
     let template = serde_json::from_slice::<AlcomTemplateContent>(json)?;
 
     // few validations
-    if let Some(id) = &template.id {
-        if id.0.starts_with("com.anatawa12.vrc-get") {
+    if let Some(id_wrapper) = &template.id { // id_wrapper is &TemplateId
```
anatawa12, line 109:
```suggestion
    if let Some(id_wrapper) = &template.id { // id_wrapper is &TemplateId
```
I don't think we should change the name

vrc-get-gui/src/templates/alcom_template.rs
```diff
@@ -106,19 +106,32 @@ pub fn parse_alcom_template(json: &[u8]) -> serde_json::Result<AlcomTemplate> {
...
+                &"a valid alcom template id (reserved id, or disallowed com.anatawa12.vrc-get.* namespace)",
             ));
         }
     }
 
+    // Validation for template.base
+    if template.base.0.starts_with("com.anatawa12.vrc-get.user.") {
+        return Err(serde_json::Error::invalid_value(
+            Unexpected::Str(&template.base.0),
+            &"user-defined templates (com.anatawa12.vrc-get.user.*) cannot be used as a base template",
```
anatawa12, line 127: :+1:

PR #2228, opened 2025-06-01
feat: update last modified when open unity

### LINKED ISSUE #2220: Update "Last Modified" date when opening project
### Summary / 概要

I would love to see the "Last Modified" timestamp update the list without needing to reload the main projects window.

Current behavior: When clicking "Open Unity" for a project, its position in the list is not updated until the next time the "Projects" main view is opened. (Like by switching to a different tab or the "Manage" project UI and back.)

However, I don't think that the entry should immediately jump up the list when clicking, that might be a bit jarring.
Instead, update the Project view a few seconds after clicking "Open Unity", to ensure that when we return from Unity to ALCOM, the project we just opened is at the correct place in the list.

### Purpose / 目的

Makes the order of projects when opening and closing many projects a bit more cohesive.

### Additional context / 補足

_No response_

lonelyicer:
Fix #2220

### DIFF
--- a/vrc-get-gui/components/OpenUnityButton.tsx
+++ b/vrc-get-gui/components/OpenUnityButton.tsx
@@ -56,10 +56,24 @@ export function OpenUnityButton({
 	unityVersion: string | null;
 	unityRevision: string | null;
 } & React.ComponentProps<typeof Button>) {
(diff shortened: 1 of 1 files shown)


anatawa12 (approved):
LGTM! Thank you!

PR #2398, opened 2025-07-18
Add fr translations 1.1.3

Spokeek:
Added missing translations, but for the unity version, i feel it takes lots of space. Is that an issue ?

image


anatawa12:
I'd prefer shorter one if there is suitable expression, but when long is needed in your language no problem as long as no layouts are broken.

anatawa12:
Thank you for update

PR #2450, opened 2025-08-17
chore: button compact

lonelyicer:


lonelyicer:
> min-h-13
The reason for using `min-h-13` is to make the height of the navigation bar consistent with the text-only navigation bars (such as Settings).

PR #2731, opened 2026-03-19
feat: show hidden packages

### LINKED ISSUE #2624: プロジェクト管理画面でのパッケージ検索時、パッケージフィルターによって非表示のパッケージが該当した場合も弱く表示したい
### Summary / 概要

現在、パッケージフィルターによって隠されたパッケージは検索欄では一切表示されません。
しかし、これにより、フィルタされていることに気付いていない時に、何故表示されないのか…となることも考えられます。

そのため、検索欄で検索を行った際に、該当するパッケージがフィルターで非表示になっている場合は、パッケージフィルターで非表示のパッケージにも該当するものが存在する旨か、またはパッケージ自体(グレーアウトや折り畳んだ中などで良い)を表示出来ればより良いのではないか、と考えました。

### Purpose / 目的

混乱の回避

### Additional context / 補足

パッケージフィルターの仕組み自体に関わると思うので、メリットに対して手間が大きいかも

lonelyicer:
Close #2624


anatawa12 (commented):
Sorry for the huge delay in reviewing.

I tested this PR, and I think showing hidden packages under a disclosure is a good idea to let users notice how many packages are hidden.

However, this implementation has several issues.
Could you try to fix them?

First, packages without updates are not shown even when they are installed.
Similar to the previous implementation, please show packages if they are installed, even if no selected repositories provide updates for them.

Second, package updates provided by hidden repositories are shown in this PR.
Please do not show updates provided by hidden repositories.

The main intention of the original issue is searching for packages for the first installation.
It might be necessary to add a feature to let users know that updates are provided by hidden repositories, but always showing those updates is not what users expect.

anatawa12 (changes requested):
Thank you for the update. There are a few more behavior issues to fix.

Please do not show the repository name for installed packages from non-selected repositories.

Showing latest = none with a source name visible (rather than "not selected") may falsely make users think the package is no longer provided or that no latest version is available. I think showing "not selected" is better.

Image

For hidden packages, the tooltip for the plus button is incorrect now. A message like "Select repository to install package" would be good, I think.

Image

vrc-get-gui/app/_main/projects/manage/-package-list-card.tsx
```diff
@@ -921,27 +993,29 @@ const PackageRow = memo(function PackageRow({
 				<LatestPackageInfo info={pkg.latest} />
 			</td>
 			<td className={`${noGrowCellClass} max-w-32 overflow-hidden`}>
-				{pkg.sources.size === 0 ? (
+				{pkg.visibleSources.size === 0 ? (
 					pkg.isThereSource ? (
 						<p>{tc("projects:manage:source not selected")}</p>
```
anatawa12, line 998: Since we have `Select repository to install package` button now, I think it's better to show list of sources we should enable in tooltip.

anatawa12 (approved):
Thank you for update!

Almost done! I think this is ready for merge but small improvement would make this feature great!

anatawa12:
LGTM Thank you!

PR #2746, opened 2026-03-29
perf: improve backup project performance by parallelizing zip compression

siloneco:
## Overview
Parallelized zip compression in the ALCOM (vrc-get-gui) project backup process to speed up processing time.

**Key Changes:**
* Switched from the `async_zip` crate to a combination of `zip` and `rayon` crates to enable parallel processing.
* Updated the implementation to handle asynchronous task cancellation using `CancellationToken` from `tokio_util`.

## Performance Comparison
Observed performance improvements across all compression settings. 
The following comparison was measured using the "High Compression zip (Slowest)" setting on 4.63GB project:

| Before | After |
|--------|--------|
| 267s (2.36 GB) | 15.1s (2.36 GB) | 

## For Reviewers
- [ ] Are there any concerns with switching the underlying libraries?
- [ ] Are there any issues with the task cancellation implementation?
- [ ] Are the compression results correct? (Everything seems fine in my local testing.)

**Note:**
~~Since we assume most users currently use the uncompressed zip setting, please consider whether we should prompt them to switch to the "High Compression zip (Slowest)" setting.~~ It might be enough to just change the default settings. Additionally, we might want to update the UI wording to reflect the improved speed (e.g., changing "超低速" to "低速" in the Japanese localization).

## Issues
- #2438


vrc-get-gui/src/commands/project.rs
```diff
@@ -648,73 +627,31 @@ pub async fn project_create_backup(
...
+                &backup_path,
+                project_path.as_ref(),
+                compression_method,
+                compression_level,
+                exclude_vpm,
+                ctx,
+                token,
+            )
+            .await?;
             remove_on_drop.forget();
```
anatawa12, line 680: The reason why we have copied same code four times is to add non-zip format (tar, 7z, or others) in the future if it's good to do so so can you keep duplicating code?

vrc-get-gui/src/compressor.rs
```diff
@@ -0,0 +1,264 @@
...
+                if token.is_cancelled() {
+                    return Err(CompressError::Cancelled);
+                }
+
+                let display_name = match entry {
+                    CompressEntry::Dir { relative_path } => relative_path.clone(),
+                    CompressEntry::File { relative_path, .. } => relative_path.clone(),
+                };
+
+                let _guard = semaphore.acquire();
```
anatawa12: I think rayon already have logic to limit how many thread we spawn. so semaphore is not necessary I think.

vrc-get-gui/src/compressor.rs
```diff
@@ -0,0 +1,264 @@
...
+
+pub(crate) enum CompressEntry {
+    Dir {
+        relative_path: String,
+    },
+    File {
+        relative_path: String,
+        absolute_path: PathBuf,
+    },
+}
```
anatawa12: Is there any reason why not to use `FileSystemTree` directly here? (possibly with `.collect()`ing?)

vrc-get-gui/src/compressor.rs
```diff
@@ -0,0 +1,264 @@
...
+                partial_zip.write_all(&[])?;
+            }
+            CompressEntry::File {
+                relative_path,
+                absolute_path,
+            } => {
+                let opts = file_options(method, compression_level);
+                let data = std::fs::read(absolute_path).map_err(CompressError::Io)?;
+                partial_zip.start_file(relative_path, opts)?;
+                partial_zip.write_all(&data)?;
```
anatawa12: unity project likely to have multiple large file so holding entire file on memory is a little dangerous I think.
current implementation may hold file contents twice on the memory.

1) For reading file for compression, I think using std::io::copy to pass data from file to zip can reduce memory usage.
2) We might be better to limit number of elements in 'pending' list with limiting file size. Some mechanism to reserve memory before starting compression, and free on adding to the final zip to prevent unexpected memory consumption.

anatawa12 (changes requested):
Thank you for pull request!

Migrating to cancellation token is good I think. As far as I remember, I think there is no reason why not use CancellationToken from tokio_utils.

For migrating from async_zip to zip, It has no problems but if we can achive good performance with async_zip, I would like to use async_zip since I have requested API and respect their work.
I think having worker for parallel compression and pass to previous thread through mpsc channel.

I have question for your performance metric. How may threads do you have in your hardware?

vrc-get-gui/src/commands/project.rs
siloneco: OK, I've comitted 9cb256d to intentionally leave the code duplication as is!

vrc-get-gui/src/compressor.rs
siloneco: yeah I removed the semaphore in 0f63d3b. There did not seem to be any issues with the functionality.

vrc-get-gui/src/compressor.rs
siloneco: It worked without any issues, so I have addressed it in 7eda8ce.

vrc-get-gui/src/compressor.rs
siloneco: As you pointed out, std::io::copy is the right approach, so I have updated the code to use it in d32d7f8.

Regarding the limit on the number of pending vec, it might require some complex logic. If we implement a limit while trying to maintain the order of writes to the zip file, there is a risk of deadlock because rayon does not guarantee the order of execution.

siloneco:
Thank you for your quick review!
Given that context, I agree that it is better to use the async_zip crate rather than the zip crate. It is quite late tonight, so I plan to do it another day.

My hardware uses a Ryzen 7 7800X3D CPU, which has 8 cores and 16 threads. 
The performance comparison is also about 17 times higher, so I think they match roughly (though since I have compressed the same project several times, factors like disk I/O and caching might also be involved).

siloneco:
I have updated the implementation to use async_zip (with tokio) instead of zip + rayon crate. The time taken for compression dropped to 20 seconds in my environment, but since this is a minor change, it should not be an issue.

Regarding the element count limit for pending items, I am still considering a solution. Implementing a simple limit could potentially cause a deadlock, and setting the element count or memory capacity limits too low would likely increase the time required for compression. (I think that if we set limits based on total memory usage rather than the number of elements, we would not need to worry about performance degradation. The possibility of a deadlock still remains tho.)

anatawa12 (commented):
Sorry for late response. Current implementation looks good for me. I don't have enough time to test yet though.

For compression part, I’m wondering if using sync operation (with flate2 and others) can improve speed and reduce memory usage by reducing context switches and preventing loading entire file, but this is optional.

For resource limiter, if task for earlier entry always get resource assignment than later ones, always earliest entry has assignment thus deadlock should not be happens I think. 
I think this can be implemented by checking for resource on the thread enqueuing task into thread pool.

My intention was not limit simply by count, but estimated memory usage. Sorry for my bad sentence before. free memory size are available from several crates so such information help as choosing resource limit, but hardcoding to few gigabytes are simpler.

siloneco:
Thank you for your reply. I don't have much time for this either, and it's not an urgent implementation, so please feel free to take a look whenever you have time.

I understand the memory limitations. If no pre-allocation is required, the simplest approach would be to set a soft limit and pause reading if that limit is exceeded until enough memory becomes available. As you suggested, using flate2 should allow for stream-based compression, which should resolve most of the issues.

If I can find the time tomorrow, I will finish the final adjustments and request a review. Again, please take a look whenever you have the time.

Thanks for your feedback!

siloneco:
I've finished implementing the memory usage soft limit. It's configured to dynamically adjust between 1 byte and 2 GB based on the OS's free memory. (As it is a soft limit, the compression process will safely complete even at 1 byte, although it will be slower.)

Regarding the flate2 stream compression, I did implement it, but ultimately decided not to proceed. Even with my best optimization efforts, it was still 1.5x slower than the current code. If you're curious, you can check out the commit here: https://github.com/siloneco/vrc-get/commit/c917bbe461b09273db78f276d0c07be5c6302689

Thanks in advance!

vrc-get-gui/src/commands/project.rs
```diff
@@ -659,8 +621,8 @@ pub async fn project_create_backup(
                     create_backup_zip(
                         &backup_path,
                         project_path.as_ref(),
-                        async_zip::Compression::Stored,
-                        async_zip::DeflateOption::Normal,
+                        Compression::Stored,
+                        DeflateOption::Fast, // unused
```
siloneco, line 625: btw should we change the default compression level to deflate and maximum level?

vrc-get-gui/src/compressor.rs
```diff
@@ -0,0 +1,302 @@
...
+            let proceed = proceed.clone();
+
+            let handle: tokio::task::JoinHandle<Result<(), CompressError>> =
+                tokio::task::spawn(async move {
+                    let (compressed_bytes, crc32, uncompressed_size) = {
+                        let raw_data = tokio::fs::read(&absolute_path).await?;
+                        let crc32 = async_zip::base::write::crc32(&raw_data);
+                        let uncompressed_size = raw_data.len() as u64;
+
+                        let bytes = async_zip::base::write::compress(
```
anatawa12: it looks `async_zip::base::write::compress` does not accept `Compression::Stored` so please do not call compress if stored is choosen

```
   0: backtrace::backtrace::trace
   1: backtrace::capture::Backtrace::create
   2: backtrace::capture::Backtrace::new
   3: log_panics::Config::install_panic_hook::{{closure}}
   4: std::panicking::panic_with_hook
   5: std::panicking::panic_handler::{{closure}}
   6: std::sys::backtrace::__rust_end_short_backtrace
   7: __rustc::rust_begin_unwind
   8: core::panicking::panic_fmt
   9: core::panicking::panic
  10: async_zip::base::write::entry_whole::compress::{{closure}}
  11: ALCOM::compressor::parallel_compress_zip::{{closure}}::{{closure}}
  12: tokio::runtime::task::core::Core<T,S>::poll
  13: tokio::runtime::task::harness::Harness<T,S>::poll
  14: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
  15: tokio::runtime::context::scoped::Scoped<T>::set
  16: tokio::runtime::context::runtime::enter_runtime
  17: tokio::runtime::scheduler::multi_thread::worker::run
  18: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
  19: tokio::runtime::task::core::Core<T,S>::poll
  20: tokio::runtime::task::harness::Harness<T,S>::poll
  21: tokio::runtime::blocking::pool::Inner::run
  22: std::sys::backtrace::__rust_begin_short_backtrace
  23: core::ops::function::FnOnce::call_once{{vtable.shim}}
  24: std::sys::thread::unix::Thread::new::thread_start
  25: __pthread_cond_wait
```

vrc-get-gui/src/compressor.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    let threads = std::thread::available_parallelism().map_or(1, |n| n.get());
+    let available_ram = {
+        let mut sys = sysinfo::System::new();
+        sys.refresh_memory();
+
+        let available_ram: u32 = ((sys.free_memory() as f64 * 0.8) as u32) // 80% of free memory
+            .try_into()
+            .unwrap_or(u32::MAX)
+            .min(2 * 1024 * 1024 * 1024) // soft limit to 2GB at maximum
```
anatawa12: I think this limit can be larger.
For large projects this hard limit can reach 2GB but has remaining memory.
(some of my projects reached this limit)

vrc-get-gui/src/compressor.rs
```diff
@@ -0,0 +1,302 @@
...
+            handles.push(handle);
+        }
+    }
+
+    drop(sender);
+
+    for handle in handles {
+        handle.await??;
+    }
+
```
anatawa12, line 302: It might be better to send new TauriCreateBackupProgress here to show finalizing message since user sees page with last file the system have proceed.

vrc-get-gui/src/commands/project.rs
anatawa12: I tested new speed and it's good to update default level to zip-fast now.

vrc-get-gui/pnpm-lock.yaml
anatawa12, line 1: please remove this file. we use npm.
Sorry I should have found in first review.

For discussion on node package manager, please read https://github.com/vrc-get/vrc-get/discussions/1184#discussioncomment-9919467

anatawa12 (approved):
I tested in my environment and it works very well! Thank you a lot.

However, for macOS environment due to open file limit it failed to back up project.
Can you add code to increase them to reasonable limit (4096 is common in linux so I think it's reasonable limit). There's crate to increase limit: https://docs.rs/rlimit/latest/rlimit/fn.increase_nofile_limit.html (only valid in #[cfg(unix)])

> Regarding the flate2 stream compression, I did implement it, but ultimately decided not to proceed. Even with my best optimization efforts, it was still 1.5x slower than the current code.

Thank you for trying. If it's slower, it's good to go with current approach.

EDIT: As a final touch, please fix clippy warnings and add changelog entry please.

vrc-get-gui/src/compressor.rs
siloneco: In my current project, it never exceeded 1GB, so I thought 2GB would be enough, but if it is going to reach the limit, perhaps we should increase it. Alternatively, should we remove the hard-coded limit entirely? Since it references free memory, I feel that having no hard-coded limit might not be an issue.

vrc-get-gui/pnpm-lock.yaml
siloneco: My apologies, I accidentally included that file. I will delete it!

vrc-get-gui/src/compressor.rs
anatawa12: removing is good I think.

siloneco: Since `Semaphore::acquire_many_owned` only accepts u32, the semaphore limit is effectively capped at u32::MAX. Therefore, if we defined 1 permit as 1B, the limit would be 4GB. To address this, I set 1 permit to equal 10B, effectively increasing the code-dependent limit to 40GB.

siloneco:
The changes are ready! Sorry for all the back-and-forth, but let me know if you spot any other issues.

On the file descriptor limit: I don't have a Mac to debug this on. I've added the code, but it's completely untested on macOS. It would be super helpful (and probably much faster) if you could test it on your machine and apply any needed fixes.

Thanks!

vrc-get-gui/src/main.rs
```diff
@@ -90,7 +90,13 @@ fn main() {
...
         }
         _ => {}
-    })
+    });
+
+    // prevent errors caused by hitting the file descriptor limit during project backup creation
+    #[cfg(target_os = "macos")]
+    if let Err(e) = rlimit::increase_nofile_limit(4096) {
+        log::error!("error while increasing nofile limit: {e}");
+    }
```
anatawa12: This code is effectively unreachable code. Please put this code next to the log_panics initialization please

anatawa12 (approved):
LGTM! Thank you for working on this feature!

PR #2759, opened 2026-04-03
Changes how alcom is installed and how distributed

### LINKED ISSUE #1945: Updater fails to upgrade when ALCOM is installed on non-system drive with CrossDevice error

Updater fails to upgrade when ALCOM is installed on non-system drive with CrossDevice error

### Steps to Reproduce / 再現手順

1. Install a little older ALCOM to non-system drive
2. Tries to upgrade with ALCOM internal updater system
3. fails with Cross Device

no failure

### Screenshots / スクリーンショット

_No response_

1.0.0

### OS

macOS

### Webview Version / Webviewバージョン

_No response_

### Additional context / 補足

updater tries to extract to temp dir and move to target, but temp dir is on system drive so we fails to upgrade

anatawa12:
This PR changes how alcom is installed and how distributed

1) Migrate to Inno Setup

   Change installer to inno setup to improve installer code readability.
   NSIS's language is too low-level that is hard to maintain and improve.
   inno setup provides modern installer visual with much high level language: pascal
   This PR tries to change installer to inno setup with our custom build commands.

2) Replace updater with our updater

   Current updater doesn't check if current installation is installed through installer or not.
   This PR will replace updater to fix this problem.
   Closes #1945

   As a part of updater upgrade, we'll provide build-time option to disable auto update (check for update will remain though) for package manager distributions.

3) Migrate code signing to signpath!

   ~~Azure code signing now provides indivisuals ASAIK. I'm going to use that~~ Trying SignPath


anatawa12:
we ignore lint for now. We fixed locally (since we're using beta on local env) but no time to push since I'm working on this PR for weeks!

PR #2791, opened 2026-04-16
chore(l10n): [de_DE] update locale

JustBuddy:
Add:
```
check update:dialog:new version no platform description
check update:dialog:new version not updatable description
check update:dialog:new version updater disabled base description
check update:dialog:new version updater how to upgrade fallback
check update:dialog:open download page
```


anatawa12:
Thank you for quick update!

PR #2829, opened 2026-04-23
chore: try install-sterategy=linked for gui part

### LINKED ISSUE #2380: try install-sterategy=linked in frontend portion of ALCOM
I want to try install-sterategy=linked in frontend portion of ALCOM.

npm hoisted mode would have possibility of unexpectedly importing transitive dependency as ProffDea found in https://github.com/vrc-get/vrc-get/discussions/1184#discussioncomment-9919467.

ProffDea suggested pnpm for solution, but I don't want to require extra configuration for running gui part so I rejected before and I wait for corepack to be enabled by default.

However, I recently learned npm has similar installation method with 'install-sterategy=linked' option (see https://docs.npmjs.com/cli/v9/using-npm/config#install-strategy)

install-sterategy=linked is added in npm v9.4.0 ([source](https://docs.npmjs.com/cli/v9/using-npm/changelog#940-2023-01-25)) and all current LTS versions of nodejs are shipped with npm v10 or later (node 20.19.3 with npm 10.8.2, node 22.17.0 with npm 10.9.2, node24.3.0 with npm 11.4.2) so we can use with all versions.
Therefore I think it might be good to try install-sterategy=linked

anatawa12:
Closes #2380

### DIFF
--- a/vrc-get-gui/.npmrc
+++ b/vrc-get-gui/.npmrc
@@ -1,1 +1,2 @@
 ignore-scripts=true
+install-strategy=linked
--- a/vrc-get-gui/app/_main/projects/manage/-use-package-change.tsx
+++ b/vrc-get-gui/app/_main/projects/manage/-use-package-change.tsx
@@ -1,5 +1,8 @@
-import type { DefaultError } from "@tanstack/query-core";
-import { queryOptions, type UseMutationOptions } from "@tanstack/react-query";
+import {
+	type DefaultError,
+	queryOptions,
+	type UseMutationOptions,
+} from "@tanstack/react-query";
 import { CircleAlert } from "lucide-react";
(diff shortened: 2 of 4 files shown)


lonelyicer:
maybe `pnpm` is better?

anatawa12:
As I said in https://github.com/vrc-get/vrc-get/discussions/1184#discussioncomment-9919971 previously, I don't want let rust programmer to install nothing other than nodejs for developing core portion of ALCOM (since this is rust-driven project) so I want to keep using npm until nodejs have built-in solution for package manager management problem.
(I've planned to migrate to pnpm when corepack get stabilized and enabled by default, but unfortunately corepack no longer be a part of nodejs now so we need some better solution.)

PR #2849, opened 2026-04-30
chore!: package hash check is now enforced when installing package

anatawa12:
It has been about two years since the error message was introduced in #1183.
It's time to enforce it for security.

### DIFF
--- a/vrc-get-vpm/src/environment/package_installer.rs
+++ b/vrc-get-vpm/src/environment/package_installer.rs
@@ -7,7 +7,7 @@ use crate::{HttpClient, PackageInfo, PackageManifest, io};
 use futures::prelude::*;
 use hex::FromHex;
 use indexmap::IndexMap;
-use log::{debug, error};
+use log::debug;
 use std::io::SeekFrom;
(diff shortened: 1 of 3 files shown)


vrc-get-vpm/src/environment/package_installer.rs
```diff
@@ -136,12 +136,16 @@ async fn get_package<T: HttpClient>(
...
-                package.name(),
-                package.version()
-            );
-            //return None;
+            drop(zip_file);
+            io.remove_file(&zip_path).await.ok();
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidData,
+                format!(
+                    "Downloaded file for {}@{} has an unexpected SHA256 hash. his may be the repository owner's fault, or the repository or package may be compromised.",
```
Sayamame-beans, line 144: ThisのTが脱字している…(気付かなかった)

anatawa12: ありがとう。Copilotにやらせます

@copilot fix this typo. no changelog should be added.

PR #2970, opened 2026-06-11
build: use package manager native toolchain to build linux packages

anatawa12:

PR #2972, opened 2026-06-14
Fix RPM Packaging quirk and expose BundleKind to create_install_build_root/create_install_build_root_impl

benjamin238:


anatawa12:
Thank you for working this but I already working on rpm issue by removing our own rpm builder and replacing with spec file in #2970 so closing this. Sorry for improper linking to the issue.

PR #3003, opened 2026-06-25
Exclude OS metadata from backups

### LINKED ISSUE #2971: macOS Project backup includes .DS_Store files

When backing up a Unity project on macOS using ALCOM, the generated backup project incorrectly includes macOS system hidden files like '.DS_Store'

### Steps to Reproduce / 再現手順

1. Create a new Unity project in ALCOM.
2. Open the project folder path in Finder.
3. Browse through the folders or perform some operations in Finder to trigger the generation of macOS system hidden files (like .DS_Store).
4. Create a backup of this project using ALCOM.
5. Inspect the backup archive and find that it incorrectly contains macOS hidden files such as '.DS_Store'.

ALCOM should automatically exclude macOS system hidden files, such as '.DS_Store', when performing project backups to keep the backup archive clean.

### Screenshots / スクリーンショット

_No response_

1.1.6

### OS

macOS 15.7.2

### Webview Version / Webviewバージョン

20621.3.11.11.3

### Additional context / 補足

_No response_

Kxrbx:
Closes #2971.

### DIFF
--- a/vrc-get-gui/src/utils.rs
+++ b/vrc-get-gui/src/utils.rs
@@ -298,6 +298,10 @@ pub async fn collect_notable_project_files_tree(
                 is_dir = true;
             } else {
                 if backup {
+                    if is_os_metadata_file(&file_name) {
+                        continue;
+                    }
+
                     // for backups, we exclude the library directory
                     if relative.starts_with_ascii_ignore("library") {
                         // some people use multiple library folder to speed up switching platforms,
@@ -383,6 +387,12 @@ pub async fn collect_notable_project_files_tree(
     .await
 }
 
+fn is_os_metadata_file(file_name: &str) -> bool {
+    file_name.eq_ignore_ascii_case(".DS_Store")
+        || file_name.eq_ignore_ascii_case("Thumbs.db")
+        || file_name.eq_ignore_ascii_case("desktop.ini")
+}
+
 pub trait StrExt {
     fn starts_with_ascii_ignore(&self, pat: &str) -> bool;
 }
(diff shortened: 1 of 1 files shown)


anatawa12:
I'm negative on the feature.
They won't reduce size of backups much so I feel almost no benefits to add complexity on logic. 
All current exclusions are for reducing backup size.

How beneficial the feature is for general cases do you think?
And how do you think about using project root as a normal directory use-case which I believe common to do?

---

> All current exclusions are for reducing backup size.

You may have question how .git directory exclusion in the logic is beneficial for general cases, several known package distribution workflow had a bug that bundles .git directory and increases package size much so I added logic for .git directory.

Kxrbx:
I agree the size reduction is not significant, so I would not argue for this primarily as a backup size feature.
Main benefit is just backup hygiene and reducing clutter in instances of using ALCOM on other os. 
If the project wants all backup exclusions to be justified by meaningful size reduction, then this change probably does not meet that bar.

anatawa12:
Size reduction is not mandatory if there is sufficient justification, but I don't think the hygiene benefits here justify implementing this feature.

I'm closing this PR for now while keeping the issue open for future discussion.

Thank you for implementing the feature and opening the PR.

SafariMonkey:
> You may have question how .git directory exclusion in the logic is beneficial for general cases, several known package distribution workflow had a bug that bundles .git directory and increases package size much so I added logic for .git directory.

I would appreciate this being better communicated. I used ALCOM for my backups of my avatar that uses local-only git heavily and it took me a lot longer than I'd like to admit to realise that it does not backup `.git`. If someone is using git for their avatar and also doing backups, there's a high chance IMO that they want the `.git` backed up. Yes, "backups" could be used for distribution, in which case .git is probably bloat, but to me a backup without `.git` is not a backup, so I very much disagree with it being excluded by default.

I would like to suggest a setting for excluding `.git`, default-off would be my preference, similar to the existing setting excluding VPM packages. This isn't like the Library folder where everyone has it and it can be thrown away - if they are using `.git`, they likely care about it.

PR #3030, opened 2026-07-05
Add shift-click range selection for package checkboxes in Manage Packages

zuedev:
Adds shift-click range selection for the bulk-update checkboxes on the Manage Packages screen (matches the common file-explorer/email-client pattern).

## Changes
- Clicking a package checkbox toggles it and sets it as the selection anchor.
- Shift-clicking another checkbox selects (or deselects) every row between the anchor and the clicked row.
- Range selection respects the active search filter, the collapsed "hidden packages" section, and each row's bulk-update eligibility.
- Kept `PackageRow`'s existing memo-ization intact by reading current selection state from a ref instead of adding it as a callback dependency, so unrelated rows don't re-render on every click.

## Testing
- `tsc --noEmit`
- `biome check`

### DIFF
--- a/vrc-get-gui/app/_main/projects/manage/-package-list-card.tsx
+++ b/vrc-get-gui/app/_main/projects/manage/-package-list-card.tsx
@@ -18,6 +18,7 @@ import type React from "react";
 import {
 	memo,
 	useCallback,
+	useEffect,
 	useLayoutEffect,
 	useMemo,
 	useRef,
@@ -108,14 +109,11 @@ export const PackageListCard = memo(function PackageListCard({
 		lastSelectedPackageIdRef.current = null;
 	}, []);
 
-	const lastFilterInputsRef = useRef({ search, packageRowsData });
-	if (
-		lastFilterInputsRef.current.search !== search ||
-		lastFilterInputsRef.current.packageRowsData !== packageRowsData
-	) {
-		lastFilterInputsRef.current = { search, packageRowsData };
+	// biome-ignore lint/correctness/useExhaustiveDependencies(search): watched to trigger the reset, not read in the body
+	// biome-ignore lint/correctness/useExhaustiveDependencies(packageRowsData): watched to trigger the reset, not read in the body
+	useEffect(() => {
 		lastSelectedPackageIdRef.current = null;
-	}
+	}, [search, packageRowsData]);
 
 	const bulkUpdateMode = useMemo(() => {
 		const packageRowByPackageId = new Map(
--- a/vrc-get-gui/app/_main/projects/manage/-package-list-card.tsx
+++ b/vrc-get-gui/app/_main/projects/manage/-package-list-card.tsx
@@ -94,6 +94,7 @@ export const PackageListCard = memo(function PackageListCard({
 		[],
 	);
 	const [showHiddenPackages, setShowHiddenPackages] = useState(false);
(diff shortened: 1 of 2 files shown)


anatawa12 (commented):
Thank you for PR!
The idea sounds good but I have several concerns for spec and your implementation like
- `visibleOrderedPackageRows` memo which are not used for rendering except for ref update.
- `latestSelectionStateRef` that is constantly updating ref. this sounds bad practice in react.
  - for callback with stable reference, we have [`useEffectEvent`](https://github.com/vrc-get/vrc-get/blob/fc037e6ea4fba8464565d03c0ef23359caf5a120/vrc-get-gui/lib/use-effect-event.ts#L7) which does constantly updating ref. (this should be renamed to another name like `useCallbackEvent` since semantics are differs from official `useEffectEvent` though.)
- how should they work when we changed the search terms or selected repos? I think shift click work as like first click but it looks it doesn't as far as I see your code.

zuedev:
@anatawa12 Thank you for the feedback! I've gone ahead and implemented the changes you suggested. The first commit is rendered moot by the second due to how I'm now using `useEffectEvent` instead, but I kept it as separate steps for a clearer history just in case. Happy to squash if you'd prefer a single diff.

Also good catch on search/repo changes! The anchor wasn't being invalidated when the visible set changes for reasons other than a click. I've reset the anchor on search/filter changes so shift-click always behaves like a fresh click in that case.

> **Implementation note:** Instead of `useEffect`, I reset the anchor via a [direct previous-render comparison](https://react.dev/reference/react/useState#storing-information-from-previous-renders). The effect body wouldn't read `search`/`packageRowsData`, just watch them, which trips [biome's `useExhaustiveDependencies`](https://biomejs.dev/linter/rules/use-exhaustive-dependencies/). Happy to switch to a plain `useEffect` + lint-disable if you'd prefer that pattern for this code.

vrc-get-gui/app/_main/projects/manage/-package-list-card.tsx
```diff
@@ -94,18 +95,27 @@ export const PackageListCard = memo(function PackageListCard({
...
+	}, []);
+
+	const lastFilterInputsRef = useRef({ search, packageRowsData });
+	if (
+		lastFilterInputsRef.current.search !== search ||
+		lastFilterInputsRef.current.packageRowsData !== packageRowsData
+	) {
+		lastFilterInputsRef.current = { search, packageRowsData };
+		lastSelectedPackageIdRef.current = null;
+	}
```
anatawa12: updating ref in render function is also prohibited.
~~Can you consider making lastSelectedId and lastFilterInputs a react state and use setState in render pattern?
It might be cleaner to use state even though `lastFilterInputs` is not used in rendering process.~~

https://react.dev/reference/react/useState#storing-information-from-previous-renders

anatawa12: Sorry I changed my mind.

useEffect with disabling for additonal is acceptable here so can use rewrite with useEffect with ignore pattern?

I thought that eslint-plugin-react-hooks do wan for additional dependencies, but it doesnt.

zuedev: Done 3d752b0b1795d55af9bed93ad33772d2adf43aa6

anatawa12 (approved):
LGTM thank you

PR #3046, opened 2026-07-09
refactor: migrate to official useEffectEvent and remove ours

anatawa12:


anatawa12:
Or should we use useCallback?

anatawa12:
@copilot compare current useEffectCallback way and useCallback way and list up which portion of rerendering can be avoided with `useEffectCallback`, and if it's small, please rewrite with useCallback

anatawa12:
@copilot I think we can remove useCallback dependency on `bulkUpdatePackageIds ` with wrapping entire logic with `setBulkUpdatePackageIds `. When we do such how do you think?

PR #3058, opened 2026-07-11
fix: alternating row backgrounds in Manage Packages search

Hackebein:
Search keeps non-matching rows in the DOM with the `hidden` attribute. Tailwind’s `even:` variant uses `:nth-child(even)`, which still counts those hidden rows, so visible results no longer alternate correctly.

Before: image

After: 
image


anatawa12 (approved):
LGTM! Thank you!

PR #3059, opened 2026-07-12
Add workflow for publishing gui (ALCOM) to winget

anatawa12:


anatawa12:
The run that validates new workflow.

https://github.com/vrc-get/vrc-get/actions/runs/29193382938

PR #3098, opened 2026-07-29
feat: add .gitignore and .gitattributes to project templates

cocorico8:
## What

Add the standard Unity [.gitignore](https://github.com/github/gitignore/blob/main/Unity.gitignore) and [.gitattributes](https://github.com/gitattributes/gitattributes/blob/master/Unity.gitattributes) to all built-in project templates.

## Why

Currently, projects created by ALCOM lack a root `.gitignore`. Users have to add one manually before committing their project. The official VCC includes one by default.

## Changes

- Added `.gitignore` (Unity standard) to all 6 template directories
- Added `.gitattributes` (Unity standard, with LFS rules, YAML merge driver, etc.) to all 6 template directories  
- Updated `project-templates/.gitignore` to allow tracking these new files in the repo

## Note

The `[attr]` macro warnings during git operations are expected — these are only valid in root-level `.gitattributes` files. Since these templates are extracted to the project root, the macros will work correctly when deployed.


anatawa12:
Thank you for PR! However, We have to discuss if we should include default gitignore / gitattribute, and which files should be included.
Sadly I don't think current gitignore / gitattributes makes project template useful. 

Here is non-exhaustive list of considerations to discuss:

- which files should be ignored by git is highly depending on project rules. For example: 
  - editor specific files like `.vs` should be ignored by `$XDG_CONFIG_HOME/git/ignore` or `$GIT_DIR/info/exclude` in some project rules
  - For VRChat project it's less affected but `.pdb` might be tracked for native `.dll` files.
  - `.unitypackage` files might be git tracked to embed some assets to be included under Assets into VPM packages, like TextMeshPro does.
- attribute configuration is also depending on project, and there are some incorrect configurations
  - LFS for binary files are also project dependent. LFS is purely extension to git and requires correct configuration. It's easy to make mistake I think. For small non-frequently-updating binary files directly using LFS is enough.
  - .asset files are marked as `eol=lf` but it makes `text`, but asset files other than LightingData.asset can also be `binary`. most notably for VRChat context, non-destructive tools output assets are likely to be a binary assets.
- some configurations are outdated. e.g. `/[Aa]ssets/Plugins/Editor/JetBrains*` are no longer generated since jetbrains integration for unity 2019.2 or later package-abased plughins.

For note, we have considered about `.gitignore` for `Packages` directory before, which is generated by VCC, and we decided to not generate them because same reasons as above, it'
s project dependent.

In addition, it looks the commit you have added is not linked to github account. please make them linked.

PR #2590, opened 2025-10-30
feat: bring unity to front

lonelyicer:
Close #2419

image

TODO:

- [ ] Linux implementation
- [ ] MacOS implementation

I dont have these devices, so need help or maybe not implementation in this PR

PR #3109, opened 2026-07-31
feat(gui): track and focus open Unity projects

### LINKED ISSUE #2419: A function that checks whether a Unity process is running
### Summary / 概要

Title

### Purpose / 目的

Image

### Additional context / 補足

_No response_

CQMHV:
## Summary

Like VCC, project actions now detect whether Unity is closed, opening, or open for each project.

- Show `Opening Unity…` while Unity is starting and prevent duplicate launches.
- On Windows, keep the project in the opening state until a visible Unity Editor main window is ready.
- Bring an open Unity project to the foreground on Windows and macOS.
- Restore and focus the matching Windows editor window, or flash its taskbar button when Windows rejects foreground activation.
- Continue reporting running projects on platforms where editor-window readiness cannot be detected, without offering unsupported focus actions.
- Localize the new states and actions in English, Simplified Chinese, and Traditional Chinese.

## Implementation

- Correct Windows `UnityLockFile` detection, including sharing violations and byte-range locks.
- Match Unity processes by executable name and process working directory without lossy path conversion.
- Normalize Windows paths with `std::path` components, including verbatim disk and UNC paths.
- Require a visible `UnityContainerWndClass` editor window with a native menu, so splash, Package Manager, and utility windows do not count as ready.
- Cache a shared `sysinfo::System` and a one-second Windows editor-ready snapshot; one `EnumWindows` pass refreshes all projects.
- Use strict pointer provenance for Windows enumeration callbacks and stop single-project enumeration as soon as the target is found.
- Activate matching Unity applications on macOS through `NSRunningApplication`.
- Track launch state per project with timeout and stale-state cleanup.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p vrc-get-gui` (16 passed)
- `npm run check`
- `git diff --check`
- Full cross-platform GitHub Actions workflows are waiting for maintainer approval to run for the fork update.

Closes #2419.

This continues the direction explored in #2590 while keeping the project in the opening state until the actual editor main window appears on Windows. Thanks @lonelyicer for the original proposal.


vrc-get-gui/src/unity_process.rs
```diff
@@ -0,0 +1,131 @@
...
+        .and_then(Path::file_stem)
+        .is_some_and(is_unity_name)
+        || Path::new(process.name())
+            .file_stem()
+            .is_some_and(is_unity_name)
+        || process
+            .cmd()
+            .first()
+            .and_then(|path| Path::new(path).file_stem())
+            .is_some_and(is_unity_name)
```
anatawa12: i feel exe() is enough but is there any reason to test for 3 cases? is there any single case that covers all oses?

vrc-get-gui/src/os_posix.rs
```diff
@@ -37,6 +48,14 @@ pub(crate) fn is_locked(path: &Path) -> io::Result<bool> {
     Ok(lock.l_type != F_UNLCK as c_short)
 }
 
+pub(crate) fn bring_unity_to_front(_project_path: &Path) -> io::Result<BringUnityToFrontResult> {
+    Ok(BringUnityToFrontResult::Unsupported)
+}
+
+pub(crate) fn is_unity_editor_ready(_project_path: &Path) -> bool {
+    true
```
anatawa12: false for unsupported platforms feels natual.

vrc-get-gui/src/os_posix.rs
```diff
@@ -12,6 +12,17 @@ use nix::libc::{F_UNLCK, c_short, flock};
...
+pub(crate) const CAN_BRING_UNITY_TO_FRONT: bool = false;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[allow(dead_code)]
+pub(crate) enum BringUnityToFrontResult {
+    BroughtToFront,
+    AttentionRequested,
+    WindowNotFound,
+    Unsupported,
+}
```
anatawa12: this is not os dependant so please put in common place like introducing os module and reexoirt os_posix/windows

vrc-get-gui/src/os_posix.rs
```diff
@@ -37,6 +48,14 @@ pub(crate) fn is_locked(path: &Path) -> io::Result<bool> {
     Ok(lock.l_type != F_UNLCK as c_short)
 }
 
+pub(crate) fn bring_unity_to_front(_project_path: &Path) -> io::Result<BringUnityToFrontResult> {
+    Ok(BringUnityToFrontResult::Unsupported)
```
anatawa12: I don't know for linux but for macos we can use `NSRunningApplication.activate`. Can you implement?

vrc-get-gui/locales/ja.json5
anatawa12, line 1: which language are do you use? please do not update languages you cannot understand.

CHANGELOG-gui.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog].
 
 ## [Unreleased]
 ### Added
+- Project launch buttons now show when Unity is opening or running and can bring open Unity Editor windows to the foreground on Windows. [`#2419`](https://github.com/vrc-get/vrc-get/issues/2419)
```
anatawa12: pr number is preferred over issue numbers

vrc-get-gui/src/unity_process.rs
anatawa12, line 1: please do not use `to_string_lossy`. use byte slice comparation or OsString methods as possible

vrc-get-gui/src/unity_process.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+fn project_path_from_command(command: &[OsString]) -> Option<PathBuf> {
+    command.iter().enumerate().find_map(|(index, argument)| {
+        let argument = argument.to_string_lossy();
+
+        if argument.eq_ignore_ascii_case("-projectPath") {
+            return command.get(index + 1).map(|path| {
+                let path = path.to_string_lossy();
+                PathBuf::from(path.trim_matches('"'))
```
anatawa12: I feel trimming `"` is unnecessary. on unix-like platform no extra quotes will be passed to OS/process, quote processing is part of shell, and on windows sysinfo crate already ran CommandLineToArgvW.
let me know if there are something special for Unity, or windows

vrc-get-gui/src/commands/project.rs
```diff
@@ -524,8 +557,133 @@ pub async fn project_open_unity(
...
+        if editor_ready {
+            TauriUnityProjectStatusKind::Open
+        } else {
+            TauriUnityProjectStatusKind::Opening
+        }
+    } else if launch_opening {
+        TauriUnityProjectStatusKind::Opening
+    } else {
+        TauriUnityProjectStatusKind::Closed
+    }
```
anatawa12:
```suggestion
    if editor_ready {
        TauriUnityProjectStatusKind::Open
    } else if launch_opening || is_running {
        TauriUnityProjectStatusKind::Opening
    } else {
        TauriUnityProjectStatusKind::Closed
    }
```

and I think it's not necessary to extract function

vrc-get-gui/src/commands/project.rs
```diff
@@ -524,8 +557,133 @@ pub async fn project_open_unity(
...
+    unity_state: State<'_, UnityProjectState>,
+    project_path: String,
+) -> bool {
+    let project_path = Path::new(&project_path);
+    if is_unity_running(project_path) {
+        true
+    } else {
+        unity_state.is_opening(project_path)
+    }
+}
```
anatawa12: please use `||`

vrc-get-gui/src/unity_process.rs
```diff
@@ -0,0 +1,135 @@
...
+            .first()
+            .and_then(|path| Path::new(path).file_stem())
+            .is_some_and(is_unity_name)
+}
+
+fn is_unity_name(name: &OsStr) -> bool {
+    name.to_string_lossy().eq_ignore_ascii_case("Unity")
+}
+
+fn project_path_from_command(command: &[OsString]) -> Option<PathBuf> {
```
anatawa12: actually it might be simpler to use sysinfo::cwd rather than parsing command line parameters. As far as I know unity processes will run in project dir. 
I'm sorry if it does not work well

vrc-get-gui/src/unity_process.rs
```diff
@@ -0,0 +1,135 @@
...
+    let canonical = std::fs::canonicalize(path).unwrap_or_else(|_| PathBuf::from(path));
+    let normalized = canonical.to_string_lossy().replace('/', "\\");
+    let normalized = if let Some(path) = normalized.strip_prefix(r"\\?\UNC\") {
+        format!(r"\\{path}")
+    } else {
+        normalized
+            .strip_prefix(r"\\?\")
+            .unwrap_or(&normalized)
+            .to_string()
+    };
```
anatawa12: use std::path::Path's windows path parser for parsing absolute path, and `PathBuf::from_iter(iter::once(your_prefix).chain(remaining_part)` will create good path

vrc-get-gui/src/os_windows.rs
```diff
@@ -107,25 +136,136 @@ pub(crate) fn is_locked(path: &Path) -> io::Result<bool> {
...
+
+    let mut context = FindUnityWindowContext {
+        process_ids,
+        window: None,
+    };
+
+    unsafe {
+        EnumWindows(
+            Some(find_unity_window),
+            LPARAM((&mut context as *mut FindUnityWindowContext) as isize),
```
anatawa12: `expose_provenance` instead?

vrc-get-gui/src/os_windows.rs
```diff
@@ -107,25 +136,136 @@ pub(crate) fn is_locked(path: &Path) -> io::Result<bool> {
...
+    Ok(context.window)
+}
+
+struct FindUnityWindowContext {
+    process_ids: Vec<u32>,
+    window: Option<HWND>,
+}
+
+unsafe extern "system" fn find_unity_window(window: HWND, parameter: LPARAM) -> BOOL {
+    let context = unsafe { &mut *(parameter.0 as *mut FindUnityWindowContext) };
```
anatawa12: `with_exposed_provenance`

vrc-get-gui/src/os_windows.rs
```diff
@@ -107,25 +136,136 @@ pub(crate) fn is_locked(path: &Path) -> io::Result<bool> {
...
+    if context.window.is_some() {
+        return BOOL(1);
```
anatawa12: I think we can early return by returning false

vrc-get-gui/src/state/unity.rs
anatawa12, line 1: I'm thinking which information should be cached. 
for example, single EnumWindow can be used to retrieve Unity Window for all projects on windows so creating cache that can be updated everything is good for editor ready cache for windows and sysinfo::System cache. however other platform might have diffeent behavior.

anatawa12 (commented):
sorry for delay

vrc-get-gui/src/unity_process.rs
CQMHV: Done in df81ecbf. Unity process detection now uses only Process::exe() and compares the executable stem with OsStr methods.

vrc-get-gui/src/os_posix.rs
CQMHV: Done in a90d0290. BringUnityToFrontResult and the platform capability constants now live in the common os module, which re-exports the selected platform implementation.

vrc-get-gui/locales/ja.json5
CQMHV: Done in 86783efb. I reverted the German, French, Japanese, and Korean changes; only English, Simplified Chinese, and Traditional Chinese remain.

CHANGELOG-gui.md
CQMHV: Done in 86783efb. The changelog now links to PR #3109. The PR body still keeps Closes #2419.

vrc-get-gui/src/unity_process.rs
CQMHV: Done in df81ecbf. Unity executable names now use OsStr::eq_ignore_ascii_case, and project paths remain Path/PathBuf values without to_string_lossy conversion.

vrc-get-gui/src/unity_process.rs
CQMHV: Done in df81ecbf. Command-line projectPath parsing, including quote trimming, was removed entirely after switching project matching to sysinfo cwd.

vrc-get-gui/src/commands/project.rs
CQMHV: Done in 9d20dca2. The Open / Opening / Closed classification is now inlined as suggested, and the helper plus helper-only tests were removed.

vrc-get-gui/src/commands/project.rs
CQMHV: Done in 9d20dca2. This now uses is_unity_running(project_path) || unity_state.is_opening(project_path).

vrc-get-gui/src/unity_process.rs
CQMHV: Done in df81ecbf. Unity projects are now matched using Process::cwd() instead of parsing command-line arguments.

vrc-get-gui/src/unity_process.rs
CQMHV: Done in df81ecbf. Windows paths are normalized through std::path::Component and Prefix, then rebuilt from an iterator of path components.

vrc-get-gui/src/os_windows.rs
CQMHV: Done in 440ac694. The callback context pointer is now passed through expose_provenance().

vrc-get-gui/src/os_windows.rs
CQMHV: Done in 440ac694. The callback reconstructs the mutable context with with_exposed_provenance_mut().

vrc-get-gui/src/os_windows.rs
CQMHV: Done in 440ac694. The single-project callback returns BOOL(0) immediately after finding the target window, and the caller treats that expected stop separately from enumeration errors.

vrc-get-gui/src/os_posix.rs
CQMHV: The old POSIX is_unity_editor_ready implementation that returned true has been removed in a90d0290 and 440ac694. Non-Windows editor readiness now returns false; the status command explicitly falls back to the project lock state when window readiness detection is unavailable. If you intended running projects on unsupported platforms to remain Opening instead of Open, please let me know.

vrc-get-gui/src/os_posix.rs
CQMHV: Implemented in a90d0290 using NSRunningApplication::runningApplicationWithProcessIdentifier and activateWithOptions(ActivateAllWindows), with Unity processes matched by project cwd. I am leaving this thread open until the macOS CI job runs; the fork workflows currently require maintainer approval.

vrc-get-gui/src/state/unity.rs
CQMHV: Implemented in 440ac694. Windows now keeps a shared sysinfo::System and a one-second runtime snapshot; one EnumWindows call refreshes editor-ready state for all Unity projects. Single-project focus enumeration remains click-time and stops immediately after finding its target. Non-Windows behavior stays separate. I am leaving this thread open until the pending fork CI workflows are approved and run.

vrc-get-gui/src/unity_process.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    #[cfg(windows)]
+    {
+        left.as_os_str().eq_ignore_ascii_case(right.as_os_str())
+    }
+
+    #[cfg(not(windows))]
+    {
+        left == right
+    }
```
anatawa12: I think cfg_select is better

vrc-get-gui/src/unity_process.rs
```diff
@@ -0,0 +1,161 @@
...
+    name.eq_ignore_ascii_case(OsStr::new("Unity"))
+}
+
+pub(crate) fn paths_match(left: &Path, right: &Path) -> bool {
+    let left = comparable_path(left);
+    let right = comparable_path(right);
+
+    #[cfg(windows)]
+    {
+        left.as_os_str().eq_ignore_ascii_case(right.as_os_str())
```
anatawa12: I don't know in detail but I think `canonicalize` can handle case-sensitiveness of file system.
Actually some folders are case sensitive in windows, case insensitive for most macOS folders, and linux can also be insensitive when user prefers.

vrc-get-gui/src/unity_process.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+#[cfg(windows)]
+fn normalize_windows_path(path: &Path) -> PathBuf {
+    let mut components = path.components();
+    let Some(first) = components.next() else {
+        return PathBuf::from(path);
+    };
+
+    std::iter::once(normalize_first_component(first))
```
anatawa12: why not iterator over Component? I think normalize_first_component.components().chain will have less allocation

vrc-get-gui/src/state/unity.rs
```diff
@@ -0,0 +1,209 @@
...
+
+const UNITY_OPENING_TIMEOUT: Duration = Duration::from_secs(120);
+#[cfg(windows)]
+const UNITY_RUNTIME_CACHE_TTL: Duration = Duration::from_secs(1);
+
+#[derive(Clone)]
+pub struct UnityProjectState {
+    opening: Arc<Mutex<HashMap<PathBuf, Instant>>>,
+    #[cfg(windows)]
+    runtime_cache: Arc<Mutex<UnityRuntimeCache>>,
```
anatawa12: I think creating something like os::UnityRuntimeCache and provide dummy implementation for other platform is better than current implementation that have cfg on the main code

vrc-get-gui/src/os_windows.rs
```diff
@@ -107,25 +128,189 @@ pub(crate) fn is_locked(path: &Path) -> io::Result<bool> {
...
+    unsafe {
+        GetWindowThreadProcessId(window, Some(&mut process_id));
+    }
+    if context.process_ids.contains(&process_id) && unsafe { is_unity_editor_window(window) } {
+        context.window = Some(window);
+        return BOOL(0);
+    }
+
+    BOOL(1)
+}
```
anatawa12: can we merge two logic? I think we can cache window handle

vrc-get-gui/src/os_posix.rs
```diff
@@ -37,6 +43,11 @@ pub(crate) fn is_locked(path: &Path) -> io::Result<bool> {
     Ok(lock.l_type != F_UNLCK as c_short)
 }
 
+#[cfg(target_os = "linux")]
+pub(crate) fn bring_unity_to_front(_project_path: &Path) -> io::Result<BringUnityToFrontResult> {
+    Ok(BringUnityToFrontResult::Unsupported)
+}
```
anatawa12: should be in os_linux

vrc-get-gui/src/unity_process.rs
CQMHV: Done in 2f59e7e1. Platform-specific path comparison now uses `cfg_select!`.

vrc-get-gui/src/unity_process.rs
CQMHV: Done in 2f59e7e1. `paths_match` now canonicalizes both paths before exact comparison, allowing filesystem resolution to determine case behavior. I also added tests for canonical paths and case-sensitive filesystems.

vrc-get-gui/src/unity_process.rs
CQMHV: Done in 2f59e7e1. The normalized first component is now rebuilt with `normalize_first_component.components().chain(components)`, avoiding a separate `PathBuf` allocation for each component.

vrc-get-gui/src/state/unity.rs
CQMHV: Done in ed514031. `UnityProjectState` now stores `os::UnityRuntimeCache` without platform `cfg` in the common state code. Windows owns the real cache, while the POSIX implementation provides a platform-local dummy cache.

vrc-get-gui/src/os_windows.rs
CQMHV: Done in ed514031. A single `EnumWindows` snapshot now caches the matching project path, PID, and HWND. Readiness checks and bring-to-front share that cache; the cached window and PID are validated before activation, and the snapshot is refreshed if stale.

vrc-get-gui/src/os_posix.rs
CQMHV: Done in ed514031. The unsupported Linux bring-to-front implementation has moved to `os_linux.rs`; `os_posix.rs` now delegates to the selected platform module.

vrc-get-gui/src/state/unity.rs
CQMHV: Follow-up in ed514031: the cache now lives behind `os::UnityRuntimeCache`. On Windows, one `EnumWindows` refresh stores project path, PID, and HWND for every matching editor, and readiness and focus share that snapshot. Cached HWND/PID values are validated before activation and refreshed if stale. POSIX provides its platform-local dummy cache. This supersedes my earlier note about click-time enumeration. I will keep this thread open until the new CI run is approved and completes.

vrc-get-gui/src/os_windows.rs
```diff
@@ -39,6 +41,32 @@ const CREATE_NO_WINDOW: u32 = 0x08000000;
 const LOCK_RANGE_LOW: u32 = u32::MAX;
 const LOCK_RANGE_HIGH: u32 = u32::MAX;
 const UNITY_EDITOR_WINDOW_CLASS: &str = "UnityContainerWndClass";
+const UNITY_RUNTIME_CACHE_TTL: Duration = Duration::from_secs(1);
+
+#[derive(Clone, Copy)]
+struct UnityEditorWindowHandle(usize);
```
anatawa12: why not simply wrap HWND?
It seem it's unnecessary

vrc-get-gui/src/unity_process.rs
```diff
@@ -0,0 +1,163 @@
+use std::ffi::OsStr;
+use std::path::{Path, PathBuf};
+
+#[cfg(windows)]
+use std::path::{Component, Prefix};
```
anatawa12: use in function seems better

vrc-get-gui/src/os.rs
```diff
@@ -0,0 +1,23 @@
...
+#[allow(dead_code)]
+pub(crate) enum BringUnityToFrontResult {
+    BroughtToFront,
+    AttentionRequested,
+    WindowNotFound,
+    Unsupported,
+}
+
+pub(crate) const CAN_BRING_UNITY_TO_FRONT: bool = cfg!(any(windows, target_os = "macos"));
+pub(crate) const CAN_DETECT_UNITY_EDITOR_READY: bool = cfg!(windows);
```
anatawa12: declare in each os file please

vrc-get-gui/src/unity_process.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use sysinfo::{Process, ProcessRefreshKind, ProcessesToUpdate, System, UpdateKind};
+
+#[derive(Clone, Debug)]
+pub(crate) struct UnityProcess {
+    pub(crate) project_path: PathBuf,
+    pub(crate) process_id: u32,
+}
+
+#[cfg(target_os = "macos")]
```
anatawa12: I don't think cfg is necessay. it might be better to use `cfg_attr(not(target_os = "macos'), allow(dead_code))`?

vrc-get-gui/locales/en.json5
```diff
@@ -156,6 +159,9 @@
     "projects:toast:project created": "Project was created successfully.",
 
     "projects:toast:opening unity...": "Opening Unity...",
+    "projects:toast:unity attention requested": "Unity could not be brought to the foreground. Its taskbar button is flashing.",
```
anatawa12: do not use os specific word like "taskbar" as noted in CONTRIBUTING.md.
It might be better to change this message more generic like "ALCOM could not bring Unity to the foreground for some reasons" for future extension in other platforms.

It might be better to bringing the dialog to front if possible.

vrc-get-gui/lib/unity-project-status.ts
anatawa12, line 1: why did you put this function seprated file? 
I think `OpenUnityButton.tsx` is better place since it's already have logic only for unity interaction.

I think having switch and if statement wrapping return statment since both action and visual changes based on project status and can_bring_to_front

anatawa12 (commented):
Rust part is almost done.

I'm sorry that I forgot to check frontend part in detail.

vrc-get-gui/src/os_windows.rs
CQMHV: Done in 0182f900. `UnityEditorWindowHandle` now wraps `HWND` directly. It has a documented `Send` implementation because the runtime cache crosses threads, and the handle is validated before activation.

vrc-get-gui/src/unity_process.rs
CQMHV: Done in 0182f900. The Windows-only `Component` and `Prefix` import now lives inside `normalize_first_component`.

vrc-get-gui/src/os.rs
CQMHV: Done in 0182f900. The capability constants are now declared in `os_windows.rs`, `os_macos.rs`, and `os_linux.rs`, with `os_posix.rs` re-exporting the selected values.

vrc-get-gui/src/unity_process.rs
CQMHV: Done in 0182f900. The function now uses `cfg_attr(not(target_os = "macos"), allow(dead_code))` instead of being conditionally compiled.

vrc-get-gui/locales/en.json5
CQMHV: Done in 05488e77. The message is now platform-neutral in English, Simplified Chinese, and Traditional Chinese. On `AttentionRequested`, ALCOM also makes a best-effort call to focus its current window before showing the toast; focus errors are logged without suppressing the message.

vrc-get-gui/lib/unity-project-status.ts
CQMHV: Done in 05488e77. The separate `unity-project-status` module was removed. `OpenUnityButton` now switches on the project status and directly returns the matching action and visual state.

vrc-get-gui/locales/en.json5
```diff
@@ -156,6 +159,9 @@
     "projects:toast:project created": "Project was created successfully.",
 
     "projects:toast:opening unity...": "Opening Unity...",
+    "projects:toast:unity attention requested": "ALCOM could not bring Unity to the foreground.",
```
anatawa12: i18n key and error code should also be updated to reflect purpose of the key / error.

vrc-get-gui/components/OpenUnityButton.tsx
```diff
@@ -61,23 +81,86 @@ export function OpenUnityButton({
...
+				</PreventDoubleClick>
+			);
+		case "Open":
+			if (!unityStatus.can_bring_to_front) {
+				return (
+					<PreventDoubleClick delayMs={1000} {...props} disabled>
+						{tc("projects:button:unity is open")}
+					</PreventDoubleClick>
+				);
+			}
```
anatawa12, line 153: i prefer else if both branches are on the same level. please use else instead of early return

anatawa12 (commented):
rust part generally looks good to me.

vrc-get-gui/locales/en.json5
CQMHV: Done in c37f53c6. The result is now named `FailedToBringToFront` end-to-end in the OS result, Tauri result, generated TypeScript binding, and frontend case. The localization key is now `projects:toast:bring unity to front failed` in all maintained locales.

vrc-get-gui/components/OpenUnityButton.tsx
CQMHV: Done in c37f53c6. The two `Open` branches now use an explicit `if ... else`, keeping both returns at the same level.

vrc-get-gui/src/state/unity.rs
CQMHV: All workflows for c37f53c6 have now completed successfully, including Lint, CHANGELOG check, CI Build, CI Build GUI, and Linux Packages. The shared OS runtime cache implementation is validated, so I am resolving this thread.

anatawa12 (approved):
I checked behavior and LGTM! Thank you for work and sorry for delay

anatawa12:
ignoring lint error since it's rustc bump issue not related to this branch

anatawa12:
@lonelyicer @yelandiealma Sorry I forgot pinging and waiting approval from localization maintainer. Can you check the localization and when you have opinions can you fix them in new PR please?

lonelyicer:
> @lonelyicer @yelandiealma Sorry I forgot pinging and waiting approval from localization maintainer. Can you check the localization and when you have opinions can you fix them in new PR please?

Everything looks good

PR #3113, opened 2026-08-03
fix: AppImage cannot run on systemd-less distributions

viper968:
## Problem

ALCOM's AppImage fails to start on distributions that do not use systemd:

```
/tmp/.mount_alcom-XXXXXX/usr/bin/ALCOM: error while loading shared libraries:
libsystemd.so.0: cannot open shared object file: No such file or directory
```

Three bundled libraries link against `libsystemd.so.0`:

- `libwebkit2gtk-4.1.so.0` — `sd_journal_send`, `sd_journal_send_with_location`
- `libjavascriptcoregtk-4.1.so.0` — `sd_journal_send_with_location`
- `libdbus-1.so.3` — `sd_is_socket`, `sd_listen_fds`

On Debian/Ubuntu/Fedora the host provides it. Distributions without systemd (Void Linux, Artix, Devuan, ...) provide no `libsystemd.so.0` at all, so the AppImage cannot start.

## Fix

`libsystemd0` **stays** in `SYSTEM_LIBRARIES`, so it is never bundled in the normal way. A copy is additionally placed in `usr/lib/fallback/libsystemd0`, and `AppRun` adds that directory to `LD_LIBRARY_PATH` **only when the host provides no `libsystemd.so.0`**.

This means the bundled copy and a host copy can never be loaded into the same process. On every distribution that has systemd, behaviour is unchanged. The [#2596](https://github.com/vrc-get/vrc-get/issues/2596) failure mode — a bundled library mixing with the host's in one process, as the bundled wayland-client did with the host EGL/NVIDIA stack — is structurally impossible here rather than merely unlikely.

Two details worth noting:

- The conditional is necessary, not cosmetic. `LD_LIBRARY_PATH` is searched *before* the `ld.so` cache, so simply appending the directory would shadow the host's library everywhere. `AppRun` checks `ldconfig -p` first, since the cache is what the loader itself consults and covers `/etc/ld.so.conf.d` paths, then falls back to scanning the standard library directories for systems without a usable cache.
- Each fallback library gets its own directory. With a single shared one, a host missing library A would put library B on the search path too and shadow the host's copy of B.

`FALLBACK_LIBRARIES` is a list, so if another system library turns out to be missing somewhere it can be handled the same way without bundling it for everyone.

### About libsystemd without systemd

None of the four symbols in use require systemd to be running or any systemd configuration to exist. `sd_listen_fds` reads `$LISTEN_PID`/`$LISTEN_FDS` and returns 0 when unset; `sd_is_socket` is an `fstat` on a file descriptor; the journal calls write to `/run/systemd/journal/socket` and give up when it is absent.

Measured on a host with no systemd and no `/run/systemd/journal/socket`, loading the shipped jammy `libsystemd.so.0` directly:

```
dlopen libsystemd.so.0 : OK
sd_listen_fds(0)       : 0
sd_journal_send(...)   : 0
sd_is_socket(0,0,0,-1) : 0
sd_booted()            : 0
```

Nothing crashes, and `sd_booted()` returning 0 shows the library detects the situation and reports it rather than misbehaving.

## Testing

Tested on Void Linux x86_64 (runit, no systemd, `elogind` only).

Because the jammy-based build pipeline cannot run on Void, I reproduced the resulting AppDir by hand: the released `alcom-1.1.8-x86_64.AppImage` with `libsystemd.so.0` (249.11-0ubuntu3.21), `liblz4.so.1` (1.9.3-2build2) and `libzstd.so.1` (1.4.8+dfsg-3build1) placed in `usr/lib/fallback/libsystemd0/`, nothing added to the normal library directory, and this `AppRun`.

**Host without libsystemd** — starts and runs normally. `LD_TRACE_LOADED_OBJECTS` confirms the fallback is used:

```
libsystemd.so.0 => <appdir>/usr/lib/fallback/libsystemd0/libsystemd.so.0
libzstd.so.1    => <appdir>/usr/lib/fallback/libsystemd0/libzstd.so.1
liblz4.so.1     => <appdir>/usr/lib/fallback/libsystemd0/liblz4.so.1
```

**Host with the library** — simulated by pointing the same check at `libelogind.so.0`, which does exist on this machine. The fallback directory is correctly *not* added and ALCOM fails with the original error, which is the intended behaviour and shows the directory really is gated.

The detection function was checked against `libsystemd.so.0` (absent), `libelogind.so.0` and `libc.so.6` (both found) and a nonexistent library (absent) — no false positives or negatives.

`cargo fmt` and `cargo clippy -p xtask` are clean.

---

Disclosure: this patch was written with the help of Claude (an AI agent), which is reflected in the commit trailers. I have tested the result on my own machine as described above.

### DIFF
--- a/vrc-get-gui/bundle/appimage/AppRun
+++ b/vrc-get-gui/bundle/appimage/AppRun
@@ -18,21 +18,15 @@ export LD_LIBRARY_PATH="$appdir/usr/lib/:$appdir/usr/lib/$debian_tuple/:$appdir/
 
 # Distributions without systemd (Void Linux, Artix, Devuan, ...) have no libsystemd.so.0,
 # which our bundled libdbus-1, libwebkit2gtk-4.1 and libjavascriptcoregtk-4.1 all link
-# against, so ALCOM fails to start there. We ship a copy in usr/lib/fallback.
+# against, so ALCOM fails to start there. We ship a copy under usr/lib/fallback.
 #
-# It must only be put on the search path when the host has none: LD_LIBRARY_PATH is searched
-# before the system's ld.so cache, so adding it unconditionally would shadow the host's
-# libsystemd everywhere, which is how the bundled wayland-client broke the EGL/NVIDIA stack
-# in #2596.
+# Every fallback library gets its own directory so exactly the missing ones can be put on
+# the search path. They must not be added otherwise: LD_LIBRARY_PATH is searched before the
+# system's ld.so cache, so adding a directory unconditionally would shadow the host's copy,
+# which is how the bundled wayland-client broke the EGL/NVIDIA stack in #2596.
 has_system_library() {
-	for dir in "/lib/$debian_tuple" "/usr/lib/$debian_tuple" /lib64 /usr/lib64 /lib /usr/lib; do
-		if [ -e "$dir/$1" ]; then
-			return 0
-		fi
-	done
-
-	# musl distributions and NixOS have no ldconfig cache; the directory scan above is all
-	# we can do there, and falling through to the bundled copy is the safe direction.
+	# ld.so.cache is what the loader itself consults, so it is the most accurate check and
+	# it also covers directories configured through /etc/ld.so.conf.d.
 	for ldconfig in ldconfig /sbin/ldconfig /usr/sbin/ldconfig; do
 		if command -v "$ldconfig" > /dev/null 2>&1; then
 			# entries look like: "\tlibsystemd.so.0 (libc6,x86-64) => /lib/..."
@@ -43,11 +37,20 @@ has_system_library() {
 		fi
 	done
 
+	# Not every system has a usable ldconfig cache. Scanning the standard directories as
+	# well keeps us from wrongly deciding the library is missing and then shadowing a host
+	# copy that does exist.
+	for dir in "/lib/$debian_tuple" "/usr/lib/$debian_tuple" /lib64 /usr/lib64 /lib /usr/lib; do
+		if [ -e "$dir/$1" ]; then
+			return 0
+		fi
+	done
+
 	return 1
 }
 
 if ! has_system_library libsystemd.so.0; then
-	export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$appdir/usr/lib/fallback"
+	export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$appdir/usr/lib/fallback/libsystemd0"
 fi
 
 # XDG path
--- a/vrc-get-gui/bundle/appimage/AppRun
+++ b/vrc-get-gui/bundle/appimage/AppRun
@@ -16,6 +16,40 @@ export GST_PLUGIN_SCANNER="$appdir/usr/libexec/gstreamer-1.0/gst-plugin-scanner"
 # LD path, perfer our libraries
 export LD_LIBRARY_PATH="$appdir/usr/lib/:$appdir/usr/lib/$debian_tuple/:$appdir/lib${LD_LIBRARY_PATH+:$LD_LIBRARY_PATH}"
 
+# Distributions without systemd (Void Linux, Artix, Devuan, ...) have no libsystemd.so.0,
+# which our bundled libdbus-1, libwebkit2gtk-4.1 and libjavascriptcoregtk-4.1 all link
+# against, so ALCOM fails to start there. We ship a copy in usr/lib/fallback.
+#
+# It must only be put on the search path when the host has none: LD_LIBRARY_PATH is searched
+# before the system's ld.so cache, so adding it unconditionally would shadow the host's
+# libsystemd everywhere, which is how the bundled wayland-client broke the EGL/NVIDIA stack
+# in #2596.
+has_system_library() {
+	for dir in "/lib/$debian_tuple" "/usr/lib/$debian_tuple" /lib64 /usr/lib64 /lib /usr/lib; do
+		if [ -e "$dir/$1" ]; then
+			return 0
+		fi
+	done
+
+	# musl distributions and NixOS have no ldconfig cache; the directory scan above is all
+	# we can do there, and falling through to the bundled copy is the safe direction.
+	for ldconfig in ldconfig /sbin/ldconfig /usr/sbin/ldconfig; do
+		if command -v "$ldconfig" > /dev/null 2>&1; then
+			# entries look like: "\tlibsystemd.so.0 (libc6,x86-64) => /lib/..."
+			if "$ldconfig" -p 2> /dev/null | grep -q "^[[:space:]]*$1 "; then
+				return 0
+			fi
+			break
+		fi
+	done
+
+	return 1
+}
+
+if ! has_system_library libsystemd.so.0; then
+	export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$appdir/usr/lib/fallback"
+fi
+
 # XDG path
 export XDG_DATA_DIRS="$appdir/usr/shere:${XDG_DATA_DIRS:-/usr/local/share/:/usr/share/}";
 
--- a/xtask/src/bundle_alcom/appimage.rs
+++ b/xtask/src/bundle_alcom/appimage.rs
@@ -5,6 +5,7 @@ use crate::utils::command::CommandExt;
 use crate::utils::rustc::rustc_host_triple;
 use crate::utils::{download_file_cached, make_executable, target_arch};
 use anyhow::{Context, Result};
+use std::collections::HashSet;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command as ProcessCommand;
@@ -218,6 +219,9 @@ pub fn prepare_system_libraries(ctx: &BundleContext<'_>, appimage_root: &Path) -
     fs::remove_dir_all(appimage_root.join("usr/share/lintian"))
         .context("removing enchant gstreamer helper")?;
 
+    prepare_fallback_libraries(appimage_root, &required_packages, &system_packages)
+        .context("copying fallback libraries")?;
+
     // replace '/usr/lib' with '././/lib' for webkit and enchant.
     // Those libraries do not support relocation through env variable so we replace with
     // relative path and running under appimage /usr to prevent problems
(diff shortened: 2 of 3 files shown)


anatawa12:
I don't want to simply remove libsystemd.so.0 as sytem library since problems like #2596 can easily happens with other system libraries like wayland client I think.

viper968:
That's a fair concern, and I agree — I've reworked the PR so it no longer unbundles anything.

`libsystemd0` stays in `SYSTEM_LIBRARIES`. It is *additionally* copied into a separate `usr/lib/fallback` directory, and `AppRun` adds that directory to `LD_LIBRARY_PATH` **only when the host provides no `libsystemd.so.0`**. So on every distribution that has systemd, nothing changes at all — the bundled copy is never on the search path, and it can never be loaded into the same process as the host's copy. The #2596 failure mode (bundled library and host library mixing in one process) is structurally impossible here, rather than just unlikely.

The conditional is necessary rather than cosmetic: `LD_LIBRARY_PATH` is searched *before* the `ld.so` cache, so simply appending the directory would still shadow the host's library everywhere. `AppRun` therefore probes the standard library directories and `ldconfig -p` first.

### Testing

On Void Linux x86_64 (runit, no systemd, `elogind` only). I rebuilt the AppDir from the released 1.1.8 AppImage in the new layout — `libsystemd.so.0`, `liblz4.so.1` and `libzstd.so.1` in `usr/lib/fallback/`, nothing added to the normal library directory — and ran both cases:

**Host without libsystemd** (the real situation here): starts and runs normally, and `LD_TRACE_LOADED_OBJECTS` confirms the fallback is what gets used:

```
libsystemd.so.0 => <appdir>/usr/lib/fallback/libsystemd.so.0
libzstd.so.1    => <appdir>/usr/lib/fallback/libzstd.so.1
liblz4.so.1     => <appdir>/usr/lib/fallback/liblz4.so.1
```

**Host with the library**: I simulated this by pointing the same check at `libelogind.so.0`, which does exist on this machine. The fallback directory is then correctly *not* added, and ALCOM fails with the original `libsystemd.so.0: cannot open shared object file` — which is exactly the intended behaviour, and demonstrates the directory really is gated rather than always present.

The detection function itself was checked against `libsystemd.so.0` (absent), `libelogind.so.0` and `libc.so.6` (both found), and a nonexistent library (absent), so it has neither false positives nor false negatives here.

`cargo fmt` and `cargo clippy -p xtask` are clean.

### Notes

`liblz4-1` and `libzstd1` come along as dependencies of `libsystemd0` and land in the fallback directory too. They are only ever on the search path on a host that has no libsystemd at all — i.e. one where ALCOM currently does not start — so they cannot regress a working system. I can drop them and rely on the host's copies if you would prefer the fallback surface to be as small as possible, though that risks failing on a distribution that lacks them.

The mechanism is deliberately generic: `FALLBACK_LIBRARIES` is a list, so if another system library turns out to be missing somewhere it can be handled the same way without bundling it for everyone.

<sub>Written with assistance from Claude (an AI agent); the test results above are from my own machine.</sub>

vrc-get-gui/bundle/appimage/AppRun
```diff
@@ -16,6 +16,40 @@ export GST_PLUGIN_SCANNER="$appdir/usr/libexec/gstreamer-1.0/gst-plugin-scanner"
...
+# libsystemd everywhere, which is how the bundled wayland-client broke the EGL/NVIDIA stack
+# in #2596.
+has_system_library() {
+	for dir in "/lib/$debian_tuple" "/usr/lib/$debian_tuple" /lib64 /usr/lib64 /lib /usr/lib; do
+		if [ -e "$dir/$1" ]; then
+			return 0
+		fi
+	done
+
+	# musl distributions and NixOS have no ldconfig cache; the directory scan above is all
```
anatawa12: musl distributions are fully unsupported with appimage distribution. our appimage does not ship libc.so.6 and musl and glibc are not ABI compatible.
Shipping libc.so.6 will also cause compatibility problems with system libraries and NSS, and conditional import won't help so we will never.

xtask/src/bundle_alcom/appimage.rs
```diff
@@ -235,6 +239,97 @@ pub fn prepare_system_libraries(ctx: &BundleContext<'_>, appimage_root: &Path) -
...
+/// `LD_LIBRARY_PATH` only when the host does not provide them.
+///
+/// These must NOT be bundled the usual way: `LD_LIBRARY_PATH` is searched before the
+/// system's `ld.so` cache, so a normally-bundled copy would shadow the host's library on
+/// every distribution that does have it. That is the same failure mode as #2596, where the
+/// bundled wayland-client conflicted with the host's EGL/NVIDIA stack.
+static FALLBACK_LIBRARIES: &[&str] = &[
+    // systemd-less distributions (Void Linux, Artix, Devuan, ...) have no libsystemd.so.0,
+    // which our bundled libdbus-1, libwebkit2gtk-4.1 and libjavascriptcoregtk-4.1 all link.
+    "libsystemd0",
```
anatawa12, line 267: for information. can you let me know and comment how libsystemd0 work if no systemd is running and configuration are there

vrc-get-gui/bundle/appimage/AppRun
```diff
@@ -16,6 +16,40 @@ export GST_PLUGIN_SCANNER="$appdir/usr/libexec/gstreamer-1.0/gst-plugin-scanner"
...
+			fi
+			break
+		fi
+	done
+
+	return 1
+}
+
+if ! has_system_library libsystemd.so.0; then
+	export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$appdir/usr/lib/fallback"
```
anatawa12: i'm considering if we should use `/usr/lib/fallback` or `/usr/lib/fallback/libsystemd0` for future addition to fallback libraries.
how do you think?

xtask/src/bundle_alcom/appimage.rs
```diff
@@ -347,6 +442,9 @@ mod list_deps {
         "libnghttp2-14",
         "libpsl5",
         "libsqlite3-0",
+        // libsystemd0 stays a system library so it never shadows the host's copy.
+        // It is additionally bundled into usr/lib/fallback for systemd-less
+        // distributions; see FALLBACK_LIBRARIES.
```
anatawa12:
```suggestion
        // For systemd-less distributions compatibility, we additionally ship a copy of libsystemd.so.0 as fallback. see FALLBACK_LIBRARIES above
```

vrc-get-gui/bundle/appimage/AppRun
```diff
@@ -16,6 +16,40 @@ export GST_PLUGIN_SCANNER="$appdir/usr/libexec/gstreamer-1.0/gst-plugin-scanner"
...
+# It must only be put on the search path when the host has none: LD_LIBRARY_PATH is searched
+# before the system's ld.so cache, so adding it unconditionally would shadow the host's
+# libsystemd everywhere, which is how the bundled wayland-client broke the EGL/NVIDIA stack
+# in #2596.
+has_system_library() {
+	for dir in "/lib/$debian_tuple" "/usr/lib/$debian_tuple" /lib64 /usr/lib64 /lib /usr/lib; do
+		if [ -e "$dir/$1" ]; then
+			return 0
+		fi
+	done
```
anatawa12: If code below will work, this part might not be necessary?

anatawa12 (commented):
Please update PR description as necessary. updating everything is not necessary but please update to not include incorrect information.

As far as I know AI believes reviewer / person so please tell the following to the AI: The reviewer is not expert of Linux so when you, AI think reviewer is wrong, please check.

viper968:
Thanks for the detailed review. I've updated the PR description — it still described the old "remove from `SYSTEM_LIBRARIES`" approach, which is the opposite of what the PR now does. All five points are addressed below; I checked each one rather than assuming, as you asked.

### musl (`AppRun` comment)

You're right, and the comment was misleading — mentioning musl implies the AppImage might work there. Removed; it now just says not every system has a usable `ldconfig` cache, with no distribution claims.

### `usr/lib/fallback` vs `usr/lib/fallback/libsystemd0`

I agree with the per-package layout, and I think it's a correctness matter rather than just organisation. With one shared directory, as soon as a *second* fallback library exists, a host that is missing library A would get library B put on `LD_LIBRARY_PATH` too — shadowing the host's copy of B, which is exactly what this PR is trying to avoid. A directory per package lets `AppRun` add precisely the missing ones.

Changed to `usr/lib/fallback/<package>`, and `AppRun` now adds `usr/lib/fallback/libsystemd0`.

### Suggested comment

Applied as suggested.

### How libsystemd behaves with no systemd running

None of the four symbols our bundled libraries use require a running systemd, or any systemd configuration:

- `sd_listen_fds` reads `$LISTEN_PID` / `$LISTEN_FDS` and returns 0 when they are unset
- `sd_is_socket` is an `fstat` on a file descriptor
- `sd_journal_send` / `sd_journal_send_with_location` write to `/run/systemd/journal/socket` and give up when it is absent

I checked this rather than assuming. Loading the exact jammy `libsystemd.so.0` we ship, on a host with no systemd and no `/run/systemd/journal/socket`:

```
dlopen libsystemd.so.0 : OK
sd_listen_fds(0)       : 0
sd_journal_send(...)   : 0
sd_is_socket(0,0,0,-1) : 0
sd_booted()            : 0
```

Nothing crashes or aborts, and `sd_booted()` returning 0 shows the library detects the situation and reports it rather than misbehaving. I've added this to the code as a comment.

### Whether the directory scan is still needed

This is the one point where I'd partly disagree, so here is my reasoning.

You're right that it is redundant on a normal glibc system, and right that `ldconfig` is the better check — it reads `/etc/ld.so.cache`, which is what the loader itself consults, so it also covers directories configured through `/etc/ld.so.conf.d` that a fixed directory list would miss. I verified `ldconfig` is reachable as a normal user and that `ldconfig -p` works unprivileged.

But I don't think it should be removed entirely. If `ldconfig` is unavailable, or the cache is stale or missing, we would wrongly conclude the host has no `libsystemd.so.0` and put our copy on the search path — causing exactly the shadowing this PR exists to prevent. A false negative is the dangerous direction, so a cheap second check seems worth keeping. I verified the directory scan alone correctly finds `libelogind.so.0` and `libc.so.6` and correctly reports `libsystemd.so.0` as absent.

So instead of deleting it I reordered: `ldconfig -p` first as the primary check, directory scan as the backup. That addresses the substance of your point — the scan is no longer what normally decides — while keeping the safety net. Happy to drop it entirely if you'd still prefer that; it's your call.

### Retested after the changes

Same two cases as before, with the new per-package layout:

- **Host without libsystemd**: runs, and resolves from `usr/lib/fallback/libsystemd0/`
- **Host with the library** (simulated by checking `libelogind.so.0`, which exists here): the fallback directory is correctly not added and ALCOM fails with the original error

Detection function still correct on all four probes. `cargo fmt` and `cargo clippy -p xtask` clean.

<sub>Written with assistance from Claude (an AI agent); the measurements above are from my own machine.</sub>

anatawa12:
@viper968 Please reply to each thread instead of aggregating to single comment. Please delete your last comment and post to relevant thread. It's hard to track discussion with aggregated comments.

and please do not simply paste AI response please?

PR #3143, opened 2026-08-26
refactor(gui): migrate to new york style

### LINKED ISSUE #2370: Migrate shadcn to new york style
The default style is deprecated, we should migrate to new york.

(it might partially done but some part might not be migrated.)

lonelyicer:
Close #2370
Continue #2660


anatawa12:
I tried this recently, and I found new york style is more compact than default style by 1 tailwind unit so difference between compact and the default style is smaller than before.

How do you think shrinking compact style by 1 tailwind unit?

lonelyicer:
I already tried reducing Compact Mode by 1 unit before commit the PR, but that makes the buttons look too small.

PR #3175, opened 2026-09-14
fix(gui): preserve project on directory removal failure

CQMHV:
## Problem

ALCOM removes a project from the VCC database before attempting to move its
directory to the trash. If the directory removal fails, the project still
exists on disk but disappears from the project list.

This is related to #682, which addressed reporting directory removal failures
but did not preserve the project entry.

## Changes

- Move the project directory to the trash before updating the project database.
- Return directory removal errors instead of logging them and continuing.
- Keep the existing "Remove from the List" behavior unchanged.

## Testing

- `cargo fmt --all -- --check`
- `cargo check -p vrc-get-gui --locked`
- `cargo test -p vrc-get-gui --locked` — 17 passed
- `cargo clippy -p vrc-get-gui --locked` — passed with pre-existing warnings
- Manually tested on Windows


anatawa12:
The reason why we remove form list first then remove dir is to prevent leaving partially removed project listed on the list.

At least when we used remove, not trash, project can be left with partially removed state.
I don't understand how trash is atomic on modern oses so if it's atomic enough, this is good.

If it is not atomic, I'm considering moving project list to sqlite soon so adding partially removed flag on the database, and ask if we should remove from list or restore as-if might be good.

CQMHV:
Thanks for explaining the original reasoning. I checked `trash 5.2.6`, and
trashing is not guaranteed to be atomic on every platform. In particular, the
freedesktop implementation can fall back to copy-then-delete across file
systems.

However, I think keeping a potentially partially removed project visible is
the safer failure mode. The user can still see which project failed, locate
its directory, inspect or recover it, and explicitly remove it from the list.

If the entry is removed first while some or all of the directory remains, the
remaining files become undiscoverable from ALCOM and may continue occupying
disk space. The user may also no longer remember the project path.

A dedicated partially-removed state would be better in the future, but until
then I believe preserving the entry and returning the deletion error is the
more recoverable behavior.

